PR #2, opened 2016-06-15
red X when putting container to itself

alagner:
Trying putting an open container to itself resulted in different indications depending on the hackmover being on or off, this fixed now to a form consistent with the original game.


DominusExult:
Thanks again. I like those little fixes that bring us more in line with the original!

PR #11, opened 2016-08-04
Fix for Draxinar's Cloth Riddle

KnightCaptain:
This small pull request is only for fixing Draxinar's fourth and final riddle. The game only accepts 14 as the answer, but the correct response is 13. Since the riddle is done as a separate function I was able to fix it without touching the rest of Draxinar's misspelled conversation. This also adds a number of Silver Seed flags.


marzojr:
Since pull request #13 has all commits this one has, I will merge it instead.

PR #12, opened 2016-08-06
Gwani cloak checks

KnightCaptain:
This builds off the prior pull request. Hopefully you can merge that first.
Updates are:
@malignantmanor 's gwaniCloakCheck is now its own function number.
Both NPCs have been added.
Includes in the main usecode.uc file were added.
The readme.txt was updated.

I also added a large number of SHAPE listings in si_shapes.uc in case they are needed later on.


content/sifixes/src/header/si/si_shapes.uc
```diff
@@ -562,15 +562,12 @@ enum npc_shapes
...
-	SHAPE_AUTOMATON_WITH_CAPE = 945,		// Isstanar
+	SHAPE_AUTOMATON_WITH_CAPE = 945,			// Isstanar
 	SHAPE_HEALER = 946,
 	SHAPE_OPHIDIAN_SOLDIER = 947,
 	SHAPE_CATMAN = 957,
 	SHAPE_FEMALE_AVATAR = 989,				// Avatar from BG per prior notes
-	SHAPE_FEMALE_MAGE_IN_BLUE_GREEN = 1015,	// Elissa
-	
-	SHAPE_KNIGHT_OF_MONITOR = 1038,			// Caladin custom artwork
-	SHAPE_PIKEMAN_DARK = 1048				// Johnson custom artwork
```
rofl0r: why were these 2 removed ?

KnightCaptain: They're not part of the original game, but rather artwork I made. I have not released them as SHP files yet, so specifying them here does no good.

KnightCaptain:
This has been checked as part of the Skullcrusher automatons update. I fixed a few compile errors so this should be good to go now. There are few header items declared twice, I hope to go back and standardize them later on, but for now both the correct and incorrect spellings will remain.

marzojr:
Since pull request #13 has all commits this one has, I will merge it instead.

PR #13, opened 2016-08-08
Skullcrusher Automatons

KnightCaptain:
This builds off the prior two pull requests to add all necessary code to bring the five automatons to life.


marzojr:
I merged this manually to simplify history, and ensure that the usecode compiles at every step of the way.

PR #14, opened 2016-11-18
SIfixes and Numbered NPCs

KnightCaptain:
Forum thread: http://exult.sourceforge.net/forum/read.php?f=1&i=1721677&t=1721677

This pull request is larger than ideal, so let me know if I need to go back and re-do this piecemeal.

1) Added Ernesto as a 'real' numbered NPC rather than a spawned monster (NPC -1), and gave him a realistic schedule. He now eats at the inn, and loiters west of the blacksmith.
2) Added code that prevents spawning him to prevent further duplicates.

Then I took a look at The Silver Seed NPCs, each of which is also a spawned monster.

3) Added Isstanar, Elissa, Surok, Tsandar, Shal, Yurel, and Draxinar as numbered NPCs, and gave each schedules so they sleep. The keep inhabitants also move around and will now sit with Isstanar for a bit, but do not eat.
4) Wrote new Usecode to strip Elissa, Shal, and Draxinar of their spell shapes when they die.
5) Added code to the Rabbit function that kicks off the game start, and prevent Karnax from triggering the creation of the SI NPCs as spawned monsters.
6) Changed three of Yurel's path eggs to Quality 15 so his existing barks can finally be seen.
7) Added a name to Isstanar's shape so he's now "automaton" rather than completely blank. Due to the way Exult does shape names, this means the entire list has to be included in the patch, it appears.
8) Fix Draxinar's shape so he's taller than the table and most things around him. He was 4,4,3 before, now he's 4,4,5 similar to the BG's red dragons.
9) Created an egg that causes Draxinar to approach the Avatar when he first enters Draxinar's area.

Unrelated changes:
10) Added the correct in-hand offsets for the Goblins at Fawn Tower, which use the shape that doesn't have them. Their weapons now appear.
11) Fixed the teleporter out of the gridiron football field, going to the intended location outside of Pothos' house/island.


rofl0r:
- it seems your later commits were made with a different email address than the earlier ones.
and the later one is not linked with your github account.
- "yes" seems to be a rather non-descriptive commit message.
- also, i personally would not accept pull requests that have merge commits in them like your first commit here, because those can add a good deal of confusion to someone trying to make sense out of the commit history (also they can be used to conceal malicious/unwilling changes).
the clean way to avoid that is basing your work on the master branch of the repo you're contributing to, but do all your work in a branch. then open a PR and once upstream merged it into its master, you just pull from them and update your master, from where you can start further work in branches.
of course as i'm not a member of the core team you can happily ignore me, unless you feel the sense behind my words.

marzojr:
I actually agree with @rofl0r; last time I did an interactive rebase on the changes to make the history simpler.

KnightCaptain:
The desktop app won't take the Gmail + tagging like it should, but accepted my username and password.
I'm not a coder professionally so I'm still learning a good bit.
What can I do to move this forward? Interactive rebase?

marzojr:
I will do the interactive rebase myself; but I will note that some things are flat out not OK, such as moving spells.uc to misc dir. I do like creating a folder for eggs, but I think that the files could have the "egg_" prefix dropped.

But next time please keep a cleaner history.

KnightCaptain:
Thank you Marzo.

KnightCaptain:
I want to improve the quality of this before I resubmit it.

PR #15, opened 2016-12-02
Teleporter Fix from Football Field

KnightCaptain:
This tiny request only makes a single map edit. It fixes the blank destination of teleporter egg out of the gridiron football field, which now goes to the intended location outside of Pothos' house/island: 2265,1919. This was an annoying issue as a player, since the only way out without cheating was to die of hunger.


marzojr:
Issues:
- Avatar's name is not the default name ("Avatar");
- Why not squash the commits?

I fixed both and merged manually.

PR #19, opened 2017-08-09
Fixes for Teleport Here and both Stockings artwork

KnightCaptain:
This fixes Sprite 07, which is the blue one, typically used when an NPC teleports to a location.

This changes frames 5, 8-10, 13, 14 to make the circles more symmetrical and to add missing columns of pixels. The change is most clear on 14 which was missing 3 columns of pixels on the left.


marzojr:
Can you add some before and after screenshots so I can see the difference?

KnightCaptain:
Okay, I've added the before-and-after PNG files for each SHP submitted. This will allow you to see all of the changes. I've also updated the readme as well.

marzojr:
Ack, I am sorry, I should have been clearer: I was asking you to put the pngs into something like imgur.com so I could see the differences more easily, definitely *not* on the pull request.

marzojr:
The idea was that I could see what the pull request does without having to be at my computer, which is mostly turned off at home these days.

DominusExult:
KC, can you remove the png from the pull request again? And show the changes in a screenshot outside of the request? I cannot see a difference in the teleport here screenshot.

marzojr:
No need for removing, I can pull it manually and remove the commits.

KnightCaptain:
My apologies, I may have misinterpreted a prior suggestion from @DominusExult that side-by-side comparison images would be useful. Since I have many more small art fixes, would it be better to close this pull and create a new one along with a matching Imgur post?

KnightCaptain:
The web console isn't helping me get this done.

DominusExult:
Yeah, sorry, I had mentioned this for your art repository https://github.com/KnightCaptain/exult_art
I still think for that a easy webviewable picture would be nice along with the shape files.

marzojr:
Just add the PNGs to imgur and link to them here on a comment, I can work on extracting the correct commits.

DominusExult:
I've finally added the two fixes directly by cherry picking from your fork. Sorry it took so long.

PR #21, opened 2017-12-02
fix no-sound-device crash on osx

alagner:
Fixes bug #1922
https://sourceforge.net/p/exult/bugs/1922/

### DIFF
--- a/audio/midi_drivers/LowLevelMidiDriver.cpp
+++ b/audio/midi_drivers/LowLevelMidiDriver.cpp
@@ -145,8 +145,8 @@ using std::string;
 using std::endl;
 
 LowLevelMidiDriver::LowLevelMidiDriver() :
-MidiDriver(), mutex(nullptr), cbmutex(nullptr),cond(nullptr),
-global_volume(255), thread(nullptr)
+MidiDriver(), mutex(0), cbmutex(0),cond(0),
+global_volume(255), thread(0)
 {
 }
 
--- a/audio/midi_drivers/LowLevelMidiDriver.cpp
+++ b/audio/midi_drivers/LowLevelMidiDriver.cpp
@@ -145,8 +145,8 @@ using std::string;
 using std::endl;
 
 LowLevelMidiDriver::LowLevelMidiDriver() :
-MidiDriver(), mutex(0), cbmutex(0),cond(0),
-global_volume(255), thread(0)
+MidiDriver(), mutex(nullptr), cbmutex(nullptr),cond(nullptr),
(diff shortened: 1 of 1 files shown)


DominusExult:
seems that the PPC build that has to use ancient gcc versions is not happy with this:
LowLevelMidiDriver.cpp: In constructor ‘LowLevelMidiDriver::LowLevelMidiDriver()’:
LowLevelMidiDriver.cpp:148: error: ‘nullptr’ was not declared in this scope

DominusExult:
just adding the thread part of the pull request is enough to fix the issue and it still compiles for ppc

alagner:
Yeah, I know, the c++11 switch must have misguided me. No problem, I can revert nullptrs back to 0s.
BTW, what GCC version is used for ppc?

DominusExult:
For ppc I have to cross-compile using the gcc 4.2 that came with xcode3 back in the days.
Wjp and me took a closer look now again and the issue was introduced by me in 38e5cb92f7b0e745b9703c01493041f0ddf7b571 when I made FMOpl the default midi driver, making a non-threaded driver the default.
See current exultlogs http://log.usecode.org/exultlog.php?log=6Dec2017

alagner:
In general: even if the midi thread is not starting, the driver behaves as if it was, and data is enqueued for it, thus making it wait forever for the data to be consumed by the thread at exit. Question is: what should be done about it? I can try to redesign the driver code, so that the data is not equeued at all, or we can stick to my simple, yet rather ugly fix. I'm flexible here, please just tell me what the preferred option is.

wjp:
I'm fairly sure the `thread` condition is not the right one. Usually messages sent by `sendComMessage` get processed by `produceSamples()` (and `playSequences()`) even if there is no `thread`. This probably isn't happening if there's no audio device at all.

alagner:
@wjp no problem, I can look it up, thanks for guidance. If there's no audio dev, question arises if the messages should ever be enqueued then?

wjp:
I think ideally a sample producing midi driver wouldn't even be started if there's no main audio device. But I haven't yet looked into how to do this, or if this would have bad side effects.

rofl0r:
probably related https://sourceforge.net/p/exult/bugs/1869/

wjp:
No, that's unrelated.

DominusExult:
Exult no longer crashes when there is no sound device AFAICT.

wench:
Recently fixed crashes with sound disabled. Guessing those fixes also fixed this.

PR #49, opened 2020-09-23
Use numeric_limits instead of magic number

alagner:


istring.cc
```diff
@@ -54,7 +55,7 @@ int strncasecmp(const char *s1, const char *s2, uint32 length) {
 }
 
 int strcasecmp(const char *s1, const char *s2) {
-	return strncasecmp(s1, s2, 2147483647);
+	return strncasecmp(s1, s2, std::numeric_limits<int32_t>::max());
```
marzojr, line 58: Hm. I have no idea why it was just not using strcasecmp instead; could you please change to just remove the third parameter and include and use strcasecmp instead?

alagner: You mean to use the library function? My guess it's not been use before because it's POSIX, not standard library.

marzojr: Both strncasecmp and strcasecmp are POSIX functions (POSIX.1-2001, according to the man page), so no difference there.

alagner: Yeah, but see the whole file. `strncasecmp` is actually implemented some lines above my fix (`istring.cc:34`). The call refers to `Pentagram::strncasecmp`, not the global one. Unless Exult's implementation can be ifdefed-away and built only on Windows.

marzojr:
I was taking a closer look at the file, and I will reimplement your solution while rewriting the functions. I just noticed that as-is, they are actually broken.

marzojr:
For some reason I never got a notification email, so sorry for the delay.

Anyway: you are correct, I forgot we reimplemented those functions. Looking over, could I trouble you to do a couple changes? Specifically:
* modify `istring.h` to change `strncasecmp` to use `size_t` instead of `uint32`;
* do the same in `istring.cc`;
* change your `std::numeric_limits` change to also use `size_t`.

`size_t` is the correct type to use here, even though we probably will not be using strings that large anytime soon.

alagner:
> For some reason I never got a notification email, so sorry for the delay.
> 
> Anyway: you are correct, I forgot we reimplemented those functions. Looking over, could I trouble you to do a couple changes? Specifically:
> 
> * modify `istring.h` to change `strncasecmp` to use `size_t` instead of `uint32`;
> * do the same in `istring.cc`;
> * change your `std::numeric_limits` change to also use `size_t`.
> 
> `size_t` is the correct type to use here, even though we probably will not be using strings that large anytime soon.

No problem with that. What about compiling those functions conditionally on Windows?

marzojr:
Compiling them conditionally on Windows would run on the issue of several places using the full qualified Pentagram namespace to refer to it (e.g., on imagewin/imagewin.cc). It would require a bigger change. If you are up for it, it would be appreciated.

alagner:
Came up with the idea of deducing the type, one place is nothing, but maybe it saves some maintenance in multiple places also. Will revert that in case it fails to build.

marzojr:
Deducing the type there is overkill, and I would rather not; the POSIX version is specified as using size_t on that parameter, and that is unlikely to ever change.

alagner:
Done. Unless you want me to squash that all into a single commit to increase readability.

marzojr:
Please squash, yes.

PR #50, opened 2020-10-06
Cpp17

alagner:
As C++17 is enabled when possible, I decided to take advantage of that fact.
Please test the build on Windows, as I didn't have the proper opportunity to do so and CI seems broken.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -136,12 +136,8 @@ fi
 AC_PROG_CXX([$cxx_compilers])
 
 AC_LANG([C++])
-AX_CXX_COMPILE_STDCXX_17([noext],[optional])
-if test "$HAVE_CXX17" = "0"; then
-	AX_CXX_COMPILE_STDCXX_14([noext],[mandatory])
-else
-	AX_CXX17_HAVE_STDFS
-fi
+AX_CXX_COMPILE_STDCXX_17([noext],[mandatory])
+AX_CXX17_HAVE_STDFS
 AC_PROG_CXXCPP
 
 AC_PROG_INSTALL
--- a/configure.ac
+++ b/configure.ac
@@ -136,12 +136,11 @@ fi
 AC_PROG_CXX([$cxx_compilers])
 
 AC_LANG([C++])
-
 AX_CXX_COMPILE_STDCXX_17([noext],[optional])
 if test "$HAVE_CXX17" = "0"; then
 	AX_CXX_COMPILE_STDCXX_14([noext],[mandatory])
 else
-	AC_CHECK_HEADERS([filesystem])
+	AX_CXX17_HAVE_STDFS
 fi
 AC_PROG_CXXCPP
 
@@ -234,50 +233,6 @@ if test x$ac_cv_func_mkstemp = xyes ; then
 	AC_DEFINE(HAVE_MKSTEMP, 1, [Have mkstemp])
 fi
 
-# std::filesystem::path breaks compilation on OSX Mojave
-# or earlier versions of Xcode, despite the fact it's there
-AC_MSG_CHECKING([if std::filesystem::path compiles fine])
-AC_TRY_COMPILE([
-#if HAVE_FILESYSTEM
-#include <filesystem>
-#endif
-],
-[
-	std::filesystem::path a{};
-],
-	ac_cv_class_fs_path=yes,
-	ac_cv_class_fs_path=no)
-AC_MSG_RESULT($ac_cv_class_fs_path)
-AM_CONDITIONAL([HAVE_FILESYSTEM], [ test "x$ac_cv_class_fs_path" = xyes ])
-if test "x$ac_cv_class_fs_path" = xyes; then
-	AC_DEFINE(USE_STDFILESYSTEM, 1, [Use std::filesystem])
-fi
-
-AC_MSG_CHECKING([if std::filesystem requires linking stdc++fs])
-AC_LINK_IFELSE(
-[AC_LANG_SOURCE([
-#if HAVE_FILESYSTEM
-#include <filesystem>
-#endif
-#include <system_error>
-
-int main() {
-	std::error_code e;
-	std::filesystem::create_directory("/dev/null", e);
-}
-])],
-	[ac_cv_link_fs_stdlib=no],
-	[ac_cv_link_fs_stdlib=yes]
-)
-if test "x$ac_cv_link_fs_stdlib" = xyes -a "x$ac_cv_class_fs_path" = xyes; then
-	AC_MSG_RESULT(yes)
-	SYSLIBS="$SYSLIBS -lstdc++fs"
-else
-	AC_MSG_RESULT(no)
-fi
-
-
-
 # do we need special X11 libraries?
 AC_MSG_CHECKING([for special X11 libraries])
 if test x$x_libraries = xNONE; then
@@ -309,7 +264,6 @@ PKG_PROG_PKG_CONFIG
 # ---------------------------------------------------------------------
 # SDL
 # ---------------------------------------------------------------------
-
 EXULT_CHECK_SDL(:,AC_MSG_ERROR([[*** SDL not found!]]))
 
 # ---------------------------------------------------------------------
@@ -630,7 +584,7 @@ else
 	AC_MSG_RESULT(no)
 fi
 
-
+SYSLIBS="$SYSLIBS $STDFSLIB"
 # ---------------------------------------------------------------------
 # Alternative directories
 # ---------------------------------------------------------------------
--- a/configure.ac
+++ b/configure.ac
@@ -140,6 +140,8 @@ AC_LANG([C++])
 AX_CXX_COMPILE_STDCXX_17([noext],[optional])
 if test "$HAVE_CXX17" = "0"; then
 	AX_CXX_COMPILE_STDCXX_14([noext],[mandatory])
+else
+	AC_CHECK_HEADERS([filesystem])
 fi
 AC_PROG_CXXCPP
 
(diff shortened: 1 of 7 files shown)


rofl0r:
using the latest and "greatest" standard isn't always the best option. for examples it precludes people that don't have a bleeding edge compiler from compiling the sources.

DominusExult:
> using the latest and "greatest" standard isn't always the best option. for examples it precludes people that don't have a bleeding edge compiler from compiling the sources.

If I'm not mistaken we are already locked on C++17 or was it C++14?

marzojr:
We are currently on 14, but I would have no issues moving to 17 because there is a lot of useful improvements.

And @rofl0r as a rule of thumb, I tend to think that if you need to compile from source to get a newer version of GCC than you can get for DOS with DJGPP (currently at 10.2), then you are better off changing distro/OS. Just saying.

rofl0r:
> And @rofl0r as a rule of thumb, I tend to think that if you need to compile from source to get a newer version of GCC than you can get for DOS

i'm using GCC 6.x, mostly because every new version compiles slower and needs more RAM. if it wasn't for C++ features required by programs, i'd still be using GCC 4.7.4 (latest version written in C)

alagner:
@rofl0r this PR should not affect you in any way, as all the C++17 features are "ifdefed". It's not my decision to make, but I'd stick with what Marzo is saying and also opt for moving to a newer standard.
BTW, if I might ask...what system do you use that you don't have binary GCC available?
@DominusExult C++17 is automatically set to "when available" in autotools.

rofl0r:
> BTW, if I might ask...what system do you use that you don't have binary GCC available?

i'm using my own build-from-source distro [sabotage linux](https://github.com/sabotage-linux/sabotage)

alagner:
@marzojr I'm not an autotools expert, can you suggest sth related to those two failing builds on Travis? It looks like it doesn't link against stdc++fs... which seems fine, because (I suppose) AC_TRY_LINK manages without that; however the whole buildingd does not manage. Is my reasoning right?

files/utils.cc
```diff
@@ -414,12 +406,10 @@ bool U7exists(
     const char *fname         // May be converted to upper-case.
 ) {
 	string name = get_system_path(fname);
-	struct stat sbuf;
-
+    std::error_code err;
```
marzojr, line 409: Inconsistent indentation here

files/utils.cc
```diff
@@ -414,12 +406,10 @@ bool U7exists(
...
 	int uppercasecount = 0;
 	do {
-		bool exists = (stat(name, &sbuf) == 0);
-		if (exists)
+        if(fs::exists(name, err))
```
marzojr, line 412: Inconsistent indentation here

files/utils.cc
```diff
@@ -440,19 +430,11 @@ int U7mkdir(
...
-	ignore_unused_variable_warning(mode);
-	return mkdir(name.c_str());
-#else
-	return mkdir(name.c_str(), mode); // Create dir. if not already there.
-#endif
+
+    std::error_code err;
+    if (fs::create_directory(name, err))
+        fs::permissions(name, fs::perms(mode), err);
+    return err.value();
```
marzojr, line 437: Inconsistent indentation here

files/utils.cc
```diff
@@ -357,25 +360,14 @@ void U7remove(
...
-	struct stat sbuf;
-
+    std::error_code err;
 	int uppercasecount = 0;
 	do {
-		bool exists = (stat(name, &sbuf) == 0);
-		if (exists) {
-			std::remove(name.c_str());
+		if (fs::exists(name, err)) {
+			fs::remove(name.c_str(), err);
```
marzojr, line 367: `.c_str()` not needed

files/utils.cc
```diff
@@ -357,25 +360,14 @@ void U7remove(
...
 		}
 	} while (base_to_uppercase(name, ++uppercasecount));
-	std::remove(name.c_str());
-#endif
+	fs::remove(name.c_str(), err);
```
marzojr, line 370: `.c_str()` not needed

marzojr:
@alagner I will have to think about the best way to do it. Especially since the current need to link to libstdc++-fs is transitional, until the symbols are merged into libstdc++.

configure.ac
```diff
@@ -140,6 +140,9 @@ AC_LANG([C++])
 AX_CXX_COMPILE_STDCXX_17([noext],[optional])
 if test "$HAVE_CXX17" = "0"; then
 	AX_CXX_COMPILE_STDCXX_14([noext],[mandatory])
+else
+	AC_CHECK_HEADERS([filesystem])
+
```
marzojr, line 145: Stray line, please delete

configure.ac
```diff
@@ -232,6 +235,47 @@ if test x$ac_cv_func_mkstemp = xyes ; then
...
+	ac_cv_link_fs_stdlib=no,
+	ac_cv_link_fs_stdlib=yes)
+if test "x$ac_cv_fs_stdlib" = xyes -a "x$ac_cv_class_fs_path" = xyes; then
+	AC_MSG_RESULT(yes)
+	SYSLIBS="$SYSLIBS -lstdc++fs"
+else
+	AC_MSG_RESULT(no)
+fi
+
+
```
marzojr, line 277: Most of those should probably be moved to a m4 file in the m4 directory to avoid polluting too much (more) configure.ac.

configure.ac
```diff
@@ -232,6 +235,47 @@ if test x$ac_cv_func_mkstemp = xyes ; then
...
+AC_MSG_CHECKING([if std::filesystem requires linking stdc++fs])
+AC_TRY_LINK([
+#if HAVE_FILESYSTEM
+#include <filesystem>
+#endif
+#include <system_error>
+],
+[
+    std::error_code err;
+    auto a = std::filesystem::status("a", err);
```
marzojr, line 266: Probably need to use a symbol that won't come inlined from the headers, but from libstdc++-fs.

configure.ac
```diff
@@ -232,6 +235,47 @@ if test x$ac_cv_func_mkstemp = xyes ; then
...
+	ac_cv_class_fs_path=yes,
+	ac_cv_class_fs_path=no)
+AC_MSG_RESULT($ac_cv_class_fs_path)
+AM_CONDITIONAL([HAVE_FILESYSTEM], [ test "x$ac_cv_class_fs_path" = xyes ])
+if test "x$ac_cv_class_fs_path" = xyes; then
+	AC_DEFINE(USE_STDFILESYSTEM, 1, [Use std::filesystem])
+fi
+
+AC_MSG_CHECKING([if std::filesystem requires linking stdc++fs])
+AC_TRY_LINK([
```
marzojr, line 258: AC_TRY_LINK is considered to be obsolete; AC_LINK_IFELSE should be used instead.

configure.ac
```diff
@@ -232,6 +235,47 @@ if test x$ac_cv_func_mkstemp = xyes ; then
...
+],
+[
+    std::error_code err;
+    auto a = std::filesystem::status("a", err);
+],
+	ac_cv_link_fs_stdlib=no,
+	ac_cv_link_fs_stdlib=yes)
+if test "x$ac_cv_fs_stdlib" = xyes -a "x$ac_cv_class_fs_path" = xyes; then
+	AC_MSG_RESULT(yes)
+	SYSLIBS="$SYSLIBS -lstdc++fs"
```
marzojr, line 272: Note that the AC_TRY_LINK is only testing linking without the library; failing to link there does not mean that the library exists, and that should probably need to be checked as well.

files/filesystem.cc
```diff
@@ -0,0 +1,87 @@
...
+#endif
+    return err.value() == 0;
+}
+
+bool remove(
+    const std::string& name,
+    std::error_code& err
+) noexcept
+{
+    err = {0, std::system_category()};
```
marzojr, line 56: Inconsistent indentation

files/filesystem.cc
```diff
@@ -0,0 +1,87 @@
...
+#if defined(_WIN32) && defined(UNICODE)
+	const char *n = name.c_str();
+	int nLen = std::strlen(n) + 1;
+	LPTSTR lpszT = (LPTSTR) alloca(nLen * 2);
+	MultiByteToWideChar(CP_ACP, 0, n, -1, lpszT, nLen);
+	if(!DeleteFile(lpszT))
+        err = {GetLastError(), std::system_category()};
+#else
+    err = {std::remove(name.c_str()), std::system_category()};
```
marzojr, line 65: Inconsistent indentation

files/filesystem.cc
```diff
@@ -0,0 +1,87 @@
...
+#endif  /* (_WIN32) && defined(UNICODE) */
+    return err.value() == 0;
```
marzojr, line 67: Inconsistent indentation

files/filesystem.cc
```diff
@@ -0,0 +1,87 @@
...
+
+void permissions(
+    const std::string& file,
+    perms prms,
+    std::error_code& err
+) noexcept
+{
+    err = {::chmod(file.c_str(), static_cast<mode_t>(prms)), std::system_category()};
+#if !defined(_WIN32)
+#else
```
marzojr, line 78: `#ifdef _WIN32`

files/filesystem.h
```diff
@@ -0,0 +1,52 @@
...
+#include <string>
+
+#if (HAVE_FILESYSTEM && USE_STDFILESYSTEM)
+#include <filesystem>
+#endif /* (HAVE_FILESYSTEM && USE_STDFILESYSTEM) */
+
+namespace fs {
+
+#if (HAVE_FILESYSTEM && USE_STDFILESYSTEM)
+using namespace std::filesystem;
```
marzojr, line 39: No `using namespace` directives in headers, please.

alagner: Note it's just "repacking" the namespace (line 36). If you're still against it I think I can fiddle with ifdefs and use `namespace fs = std::filesystem;` instead.

files/filesystem.cc
alagner: I guess I've misplaced the whole block there. Should be fixed now.
I wonder if those system-specific call should be replaced with whole different implementation files, it might be easier to maintain this way.

files/filesystem.h
marzojr: Using `namespace fs = std::filesystem;` is the better option, yes. But note that you do not even use anything from `std::filesystem` in the header, so you can just move that line to the .cc instead.

files/filesystem.cc
marzojr: I do have plans of creating something like that, and remove a ton of `#ifdef`s and similar from the normal code.

files/filesystem.cc
```diff
@@ -73,14 +73,14 @@ void permissions(
...
-    //do nothing
-    ignore_unused_variable_warning(file);
-    ignore_unused_variable_warning(prms);
-    ignore_unused_variable_warning(err);
-#endif /* !defined(_WIN32) */
+#ifdef _WIN32
+//do nothing
+	ignore_unused_variable_warning(file);
+	ignore_unused_variable_warning(prms);
+	ignore_unused_variable_warning(err);
```
marzojr, line 80: These 3 can be combined in the same line, as ignore_unused_variable_warning accepts an arbitrary number of arguments.

files/filesystem.h
alagner: Yeah, but note that filesystem.cc is not compiled if filesystem header from stdlib is found. But I guess I could condtionally include that in `utils.cc` instead.

configure.ac
alagner: @marzojr sure, but which one? Theoretically this should be `filesystem::path`... but well apparently it's not. Seems like it's going to be try-and-error until it works. If it works.

alagner: Turns out that I made a mistake in variable naming, shame on me. -_-
@marzojr I made it fail to compile (as it should), now all the platforms on Travis build fine. Please let me know if the way it is fine.

alagner:
@marzojr how much should I care about CodeFactor complaining about `chmod`? No problem with switching to `fchmod` as it suggests, though I'm trying to reason if this is that much of a threat...

configure.ac
alagner: You've meant - succeeding to link?
Either way, redone that, with reinventing the wheel a bit, due to C++ name mangling...

wench:
Exult is now C++17. This pull request is now unnecessary.

PR #51, opened 2020-11-01
Use derivation instead of macro to prevent copying

alagner:
Solution per analogiam to boost::noncopyable, see http://bajamircea.github.io/coding/cpp/2017/02/22/noncopyable-adl.html

### DIFF
--- a/headers/exceptions.h
+++ b/headers/exceptions.h
@@ -55,21 +55,16 @@ public:
 	}
 };
 
-/*
- *  Classes which should not be replicatable throw an replication_exception
- */
-
-class replication_exception : public exult_exception {
-public:
-	explicit replication_exception(std::string what_arg): exult_exception(std::move(what_arg)) {  }
+//Per analogiam to boost::noncopyable
+class nonreplicatable {
+protected:
+	constexpr nonreplicatable() = default;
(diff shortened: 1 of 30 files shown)


rofl0r:
does this fix an actual issue, or is it just cosmetics?
if the latter, there are lots of bugs in the codebase waiting to be fixed, maybe that would be a more productive use of your (and the reviewer's) time.

DominusExult:
No, it's fine. Judging by all the work Marzo invested to keep warnings low, contributions like this are welcome.
We may decide not to do this but, at least for me, it's an interesting point.

headers/exceptions.h
```diff
@@ -55,22 +55,17 @@ class quit_exception : public exult_exception {
...
+//Per analogiam to boost::noncopyable
+class nonreplicatable {
+protected:
+	constexpr nonreplicatable() = default;
+	nonreplicatable(const nonreplicatable&) = delete;
+	nonreplicatable(nonreplicatable&&) = delete;
+	nonreplicatable& operator=(const nonreplicatable&) = delete;
+	nonreplicatable& operator=(nonreplicatable&&) = delete;
+	~nonreplicatable() = default;
 };
```
marzojr, line 67: Nice catch, forgot to remove this.

PR #52, opened 2020-11-02
Store Text_effect in std::list

alagner:
I thought it might be reasonable to simplify the code by using STL instead of hand-made linked lists.
On top of that, I have added some RAII - hopefully - and simplified the code even more.
Special_effects to go, they are a logical next step, but I prefer to keep those PRs separate for clarity (and I'm certain there's at least one more place in the code with hand-made lists).
If you prefer it organized in any other way - I'm open for suggestions.

As for the code itself - lambda capture is a matter of taste, I decided to do it by value as raw pointers are dealt here with.
What I also do necessarily like about readability is the `remove`/`remove_if` member call on list, and I was a bit tempted to change that to standard erase/remove idiom (as done e.g. with `vector`)...

### DIFF
--- a/effects.cc
+++ b/effects.cc
@@ -280,9 +280,8 @@ void Effects_manager::paint(
 
 void Effects_manager::paint_text(
 ) {
-	std::for_each(texts.begin(), texts.end(),
-			[](auto& txt) { txt->paint(); }
-	);
+	for (auto& txt : texts)
+		txt->paint();
 }
 
 /**
@@ -290,9 +289,8 @@ void Effects_manager::paint_text(
  */
 void Effects_manager::update_dirty_text(
 ) {
-	std::for_each(texts.begin(), texts.end(),
-			[](auto& txt) { txt->update_dirty(); }
-	);
+	for (auto& txt : texts)
+		txt->update_dirty();
 }
 /**
  *  Some special effects may not need painting.
--- a/effects.cc
+++ b/effects.cc
@@ -75,9 +75,10 @@ void Effects_manager::add_text(
 	if (!msg)           // Happens with edited games.
 		return;
 	// Don't duplicate for item.
-	for (Text_effect *each = texts; each; each = each->next)
-		if (each->is_text(item))
-			return;     // Already have text on this.
+	if (std::find_if(texts.cbegin(), texts.cend(),
+		[&](const auto& el) { return el->is_text(item); }) != texts.cend())
+		return; // Already have text on this.
+
 	auto *txt = new Text_effect(msg, item);
 //	txt->paint(this);        // Draw it.
 //	painted = 1;
@@ -135,11 +136,7 @@ void Effects_manager::add_effect(
 void Effects_manager::add_text_effect(
     Text_effect *effect
 ) {
-	effect->next = texts;       // Insert into chain.
-	effect->prev = nullptr;
-	if (effect->next)
-		effect->next->prev = effect;
-	texts = effect;
+	texts.emplace_front(effect);
 }
 
 /**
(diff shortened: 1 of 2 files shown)


effects.cc
```diff
@@ -91,14 +91,17 @@ void Effects_manager::add_text(
  *  @param y        y coord. on screen.
  */
 
-void Effects_manager::add_text(
+inline void Effects_manager::add_text(
```
jasonaowen, line 94: It looks like `add_text` is referred to in other places, and making the function `inline` breaks the build. From the [failed build log](https://travis-ci.org/github/exult/exult/jobs/740989003#L2966):

```
/home/travis/build/exult/exult/gamewin.cc:2122: undefined reference to `Effects_manager::add_text(char const*, int, int)'
```

I was able to reproduce this locally, and undoing this change fixed the problem.

```suggestion
void Effects_manager::add_text(
```

alagner: > It looks like `add_text` is referred to in other places, and making the function `inline` breaks the build. From the [failed build log](https://travis-ci.org/github/exult/exult/jobs/740989003#L2966):
> 
> ```
> /home/travis/build/exult/exult/gamewin.cc:2122: undefined reference to `Effects_manager::add_text(char const*, int, int)'
> ```
> 
> I was able to reproduce this locally, and undoing this change fixed the problem.

True, although in such case I wonder if moving the one-liner to header file to keep it inline wouldn't be a better option.
I will check it later today.

marzojr: Moving itt to the header would require including a bunch of headers. Better to just remove the inline.

marzojr (changes requested):

effects.cc
```diff
@@ -201,18 +183,14 @@ void Effects_manager::remove_text_effect(
 void Effects_manager::remove_all_effects(
     bool repaint
 ) {
-	if (!effects && !texts)
+	if (!effects && texts.empty())
 		return;
 	while (effects) {
```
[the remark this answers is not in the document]
alagner: Of course!
This is implied, I just prefer to do it in smaller batches, maybe even separate MR, but it's already work in progress :)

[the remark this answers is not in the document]
alagner: @marzojr side remark: I wonder if `texts` actually should be a list; why not make it a set of some sort? Then checking for duplicates happens automagically.
I have some doubt related to `void add_text(const char *msg, int x, int y);` when it comes to `set` but at the same time I doubt there would be collisions.

effects.h
```diff
@@ -47,17 +49,16 @@ using Game_object_weak = std::weak_ptr<Game_object>;
 class Effects_manager {
 	Game_window *gwin;      // Handy pointer.
 	Special_effect *effects;    // Sprite effects, projectiles, etc.
-	Text_effect *texts;     // Text snippets.
+	std::list<std::unique_ptr<Text_effect>> texts;     // Text snippets.
 public:
-	Effects_manager(Game_window *g) : gwin(g), effects(nullptr), texts(nullptr)
+	Effects_manager(Game_window *g) : gwin(g), effects(nullptr), texts{}
```
marzojr, line 54: I missed this before, but it is better to remove the trivial initializer for `texts` altogether, as the default constructor for list is called anyway.

effects.cc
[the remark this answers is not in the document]
marzojr: I thought of suggesting a set (or unordered_set) as well. The drawback is that it would change the ordering of displayed text. Currently, and with your changes, the newest text is drawn first, with each subsequent text going on top of all those before; with the set, the object address in RAM would set the ordering, making it basically random. In the end, at least for now, I think keeping the list is the better option.

effects.h
alagner: That might my MISRA bias ;)
Anyway, fixed.

effects.cc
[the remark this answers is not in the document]
alagner: As for `!effects`, see https://github.com/alagner/exult/tree/list_specials
I will create a PR when I'm done with that, unless you want to have it all in one place?
I'm fine with both, though be warned, there's way more code in the latter case; due to that I would not recommend reviewing it both changes as a whole, but your call in the end.

PR #53, opened 2020-11-03
Effects_manager: store effects is std::list

alagner:
Replace 2nd handcrafted linked list with std::list and take advantage of RAII.

### DIFF
--- a/effects.cc
+++ b/effects.cc
@@ -173,10 +173,7 @@ void Effects_manager::remove_all_effects(
 ) {
 	if (effects.empty() && texts.empty())
 		return;
-	auto ef = effects.begin();
-	while(ef != effects.end()) {
-		remove_effect((*(ef++)).get());
-	}
+	effects.clear();
 	texts.clear();
 	if (repaint)
 		gwin->paint();      // Just paint whole screen.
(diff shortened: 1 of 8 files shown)


effects.cc
```diff
@@ -216,15 +200,11 @@ void Effects_manager::remove_weather_effects(
...
-		// See if we're far enough away.
-		if (each->is_weather() && (!dist ||
-		                           static_cast<Weather_effect *>(each)->out_of_range(apos, dist)))
-			remove_effect(each);
-		each = next;
-	}
+	effects.remove_if([&](const auto& ef) {
+		return ef->is_weather() &&
+	           (!dist ||
+				static_cast<Weather_effect*>(ef.get())->out_of_range(apos, dist));
```
marzojr, line 206: Hm. It is not related to your code, but I dislike this static_cast. I am wondering if there should be a virtual method for effects that returns false for non-weather effects, and checks for out-of-range for weather effects, allowing this whole lambda body to be just something like `return ef->is_out_of_range_weather(apos, dist);`.

effects.cc
```diff
@@ -251,17 +226,35 @@ void Effects_manager::remove_usecode_lightning(
...
+/**
+ * Destructor
+ * Takes advantage of the fact that gwin
+ * is stored.
+ */
+Special_effect::~Special_effect(
+)
+{
+	if (in_queue())
+		gwin->get_tqueue()->remove(this);
```
marzojr, line 257: I am half-wondering if this code should not be pushed all the way to the Time_sensitive base class; it does not make sense to have to manually remove each and every Time_sensitive by hand when they are deleted. But this can be for a future set of commits, this is just me musing out loud.

effects.cc
alagner: I'm not a big fan of pseudo-dynamic cast either.
Easy solution would be to just return -1 from `get_number` on non-weather effects (but it might carry some other logic someplace else, I haven't done any thorough analysis here) and refactor this whole interface either to the form you're mentioning or similar, e.g. I'm personally a fan of returning an empty optional in such situations, and this should be doable once C++17 is enabled.

Option number 2 - a bit more complicated, but might be worth a shot: store specific effects in separate data structures, e.g. weather, usecode light, others + one large list for storing just observer pointers to execute paint in the right order.
Simpler interface (no searching etc.) vs more funky adding/removal logic.
Whichever you prefer, I'd opt for the latter.

effects.cc
alagner: I'll have a look at it.

PR #56, opened 2020-11-26
Fix memleak on MacOS

alagner:
MacOS APIs with `Create` or `Copy` in their name give ownership, thus requiring to call CFRelease on them. This commit does that using RAII idiom.

### DIFF
--- a/files/utils.cc
+++ b/files/utils.cc
@@ -641,16 +641,26 @@ string Get_home() {
 	return home_dir;
 }
 
+#if defined(MACOSX) || defined(__IPHONEOS__)
+struct CFDeleter
+{
+	void operator()(CFTypeRef ptr) const
+	{
+		if (ptr) CFRelease(ptr);
+	}
+};
+#endif
+
 void setup_data_dir(
     const std::string &data_path,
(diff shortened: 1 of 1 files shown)


DominusExult:
Thank you. Funnily I ran the Xcode analyzer the same day and it complained about this one

alagner:
Telepathy ;) @DominusExult speaking about analyzer (I guess you mean Instruments/Leaks?) I wonder if it could be used with this bug (if it's still valid, I don't know ;) ): https://sourceforge.net/p/exult/bugs/2027/
This would simply mean waiting for the bug to manifest while gathering data using Instruments, maybe its output can show sth valueable.
Another option - should this bug reproduce on Linux - would be compiling with address sanitizer.
Unfortunately, I did have no luck of reproducing it neither on mac nor Linux.

DominusExult:
The Analyzer can be run in xcode when you open the iOS Xcode project and in the menu hit -> Product -> Analyze.

But good point on trying to replicate that bug via Instruments. I *think* this kind of bug is really fixed now but it can't hurt to try and let it run for a while...

PR #61, opened 2020-12-28
[1.6.x branch] Fix compile for wuc.cc on Xcode12

mitchblank:
The 1.6 release fails to complie on Xcode 12.1 because of "byte" conflicting with "std::byte".  I noticed that this was already
renamed "byteval" in the master branch (part of the larger set of changes in d7998ecace00daa0469494377f916db386396406). I thought it would be useful to backport this to the v1.6.x branch as well.

I don't know anything about exult's release cycle or if there is going to be any more releases from 1.6.x but I am just trying to get the Homebrew formula for exult building again.


marzojr:
Thanks!

Issue #60, opened 2020-12-26
Removal of X11 dependency in Delay and Server_delay


Dragon-Baroque:
Proposed a new branch "exult_x11_dependencies" on Dragon-Baroque/exult.

It removes the ConnectionNumber / xfd dependency of Exult over X11 in Delay() ( gumps/gump_util.h ) and Server_delay ( server/server.cc ) using either pure SDL cross platform ( Delay ) or SDL  + UNIX select without xfd ( Server_delay ).

DominusExult:
Just tested and it works fine on macOS

Dragon-Baroque:
The Travis build of the pull request failed because of merge conflicts between exult/master and exult_x11_dependencies.
I merged master into exult_x11_dependencies and fixed conflicts. Then pushed the new exult_x11_dependencies.
The Travis build then passed.
Is this the right way to do that ? Should I have attempted a rebase of exult_x11_dependencies on master instead ?

marzojr:
I would prefer a rebase to master and force-push. This leaves a cleaner history.

Dragon-Baroque:
Branches exult_x11_dependencies ( and exult_user_events ) rebased to master.

PR #66, opened 2021-01-09
Exult x11 dependencies

marzojr:
Fixes issue #60.

### DIFF
--- a/Exult_Removal_of_X11_Connection_Number.md
+++ b/Exult_Removal_of_X11_Connection_Number.md
@@ -1,0 +1,60 @@
+# Removal of X11 ConnectionNumber xfd in Exult
+December 2020 by Dragon Baroque.
+
+## Analysis
+
+The X11 `ConnectionNumber` is a Unix file descriptor used deep in the SDL Event Loop to be notified of incoming X11 User events.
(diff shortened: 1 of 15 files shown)


DominusExult:
Some pointers for the last commit that did some macosx configure.ac and exult.cc things.
After thinking a bit about it, this needs to be done differently:
The SDL_DROPFILE on macOS was a big struggle to get working. It does work without problem when you still use X11, but the native part needed an SDL2 fix that was accepted into their source just a few days ago.

So the whole macosx configure.ac needs to check:
- if  --enable-exult-studio-support on Macosx check SDL2 version
- if SDL2 < 2.0.15 force the previous --enable-macosx-studio-support that adds XWIN to Exult + Studio 
    (maybe rename --enable-macosx-studio-support to --enable-macosx-studio-**x11**-support)
- if SDL2 > 2.0.15 enable Studio support

and in exult.cc:705
the #elif defined(MACOSX) && defined(USE_EXULTSTUDIO)...#endif needs to be changed to be:
```
#elif defined(MACOSX) && defined(USE_EXULTSTUDIO) && defined (XWIN)
SDL_putenv("SDL_VIDEODRIVER=x11");
#endif
```

PR #76, opened 2021-02-02
Using external libmt32emu

marzojr:
Instead of carrying a copy of the code in Exult source and having to manually update.


marzojr:
Merging and adding a few fixes after.

PR #81, opened 2021-02-07
Fix Drop Target file name recognition

Dragon-Baroque:
Commit 1 of 1 :
    shapes/u7drag.cc
      Shapes have different names and name lengths : do no use static,
      Drop Targets in Linux receive the URL minus the file:// prefix,
      Recognition test handled in reverse False vs True.

### DIFF
--- a/shapes/u7drag.cc
+++ b/shapes/u7drag.cc
@@ -31,18 +31,27 @@ Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.
 
 #include <cassert>
 
-#define FILE_URI "file:///"
+#define FILE_URI "file://"
+static const size_t FileUriLen = strlen(FILE_URI);
 
 /*
  *  Check and skip the URI introducer.
  */
 static bool verify_is_uri(
     const unsigned char *&data,
-	const char *searched_uri
+    const char *searched_uri
 ) {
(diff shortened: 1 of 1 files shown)


shapes/u7drag.cc
```diff
@@ -31,20 +31,28 @@ Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.
...
+	if (memcmp(data,
+	           searched_uri,
+	           urilen      ) == 0) {
 		data += urilen;
 		return true;
 	}
+	if (memcmp(data,
+	           searched_uri + strlen(FILE_URI),
+	           urilen       - strlen(FILE_URI)) == 0) {
+		data += urilen  - strlen(FILE_URI);
```
marzojr, line 53: Can you create a `const size_t` variable for result of the strlen call and use in these 3 locations?

Dragon-Baroque: I shall immediately.
But don't you think that any reasonable compiler would compute the constant once ?

marzojr: Almost certainly; but it makes for better code anyway.

Dragon-Baroque: Replacement force pushed.

marzojr:
Thanks.

Issue #71, opened 2021-01-25
A few ideas for ES UI improvements


marzojr:
Kind of a reminder to myself.

In terminal:
- [ ] When started from Exult on Windows, Exult should connect stdout and stderr from ES to the terminal.
- [ ] ES spams "Error sending to server" when it is not connected to Exult.

In main window:
- [ ] A message should be displayed that NPCs can only be viewed/edited in conjunction with Exult.

In Shape Info window:

- [ ] Weapon tab: family and projectile should show the shape in question when 'Specific shape' is selected.
- [ ] Ammo tab: family and sprite should show the shape in question when 'Specific shape' is selected.
- [ ] Animation tab: maybe add an animation preview? Might be more trouble than it is worth.
- [ ] NPC Paperdoll tab: it should display the composite paperdoll from head, arms, and body (see note below).
- [ ] SFX tab: there should be a play button which causes Exult to play the SFX (or an error message if Exult is not connected).
- [ ] Content Rules tab: it should display either the shape in the Shape field, or nothing for all shapes.
- [ ] Frame Flags tab: it should display the frame selected in the Frame field, or nothing for all frames.
- [ ] Frame HPs tab: it should display the frame selected in the Frame field, or nothing for all frames.
- [ ] Frame Names tab: it should display the frame selected in the Frame field, or nothing for all frames.
- [ ] Frame Usecode tab: it should display the frame selected in the Frame field, or nothing for all frames.
- [ ] Obj Paperdoll tab: it should display the frame selected in the Frame field, or nothing for all frames. Additionally, it should display the paperdoll shape and frames selected in the Paperdoll fields (see note below).
- [ ] Shape Brightness tab: it should display the frame selected in the Frame field, or nothing for all frames.
- [ ] Warmth tab: it should display the frame selected in the Frame field, or nothing for all frames.

In NPC window:
- [ ] Maybe there should be a button to show the NPC in the map.
- [ ] Flags tab: there should be a way to edit/display the current polymorph shape.
- [ ] Schedules tab: there should be a button to show the location of the schedule on the map, and to set the position from a click on Exult.

in Egg window:
- [ ] When an egg is selected in Exult, its range should be displayed in mapedit mode whenever it makes sense (party/avatar near, avatar far); and it should be updated when distance is changed in ES. 
- [ ] Jubebox, Sound Effect, and Voice tabs: there should be a button to play the music/SFX/voice in Exult.
- [ ] Missile tab: the missile sprite should be displayed.
[...]

PR #95, opened 2021-03-18
Studio : Provide and use a new C++ class "ShapeSingle" for single Shape display

Dragon-Baroque:
Contrarily to the discussion in issue #71, the code compiles and runs on Linux and on Windows, I figured out what the Drag and Drop WIN32 problem was and fixed it.

By the way, this highlighted that I should have had more compilation failures. Indeed, the three stand alone shape displays of the Shape Editor [ Gump, Body and Explosion ] claim to be Drop targets ( enable_drop ) but do not use the WIN32 path.

### DIFF
--- a/mapedit/npcedit.cc
+++ b/mapedit/npcedit.cc
@@ -32,6 +32,7 @@ Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.
 #include "servemsg.h"
 #include "shapedraw.h"
 #include "shapefile.h"
+#include "shapevga.h"
 #include "studio.h"
 #include "u7drag.h"
 #include "utils.h"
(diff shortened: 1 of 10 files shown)


mapedit/shapedraw.cc
```diff
@@ -151,7 +151,7 @@ void Shape_draw::draw_shape_centered(
 	}
 	else {
 		draw_shape(shape, (winw -  shape->get_width()) / 2,
-	                          (winh - shape->get_height()) / 2);
+	                      (winh - shape->get_height()) / 2);
```
marzojr, line 154: Introducing whitespace error here.

mapedit/npcedit.cc
```diff
@@ -466,7 +402,13 @@ void ExultStudio::init_new_npc(
 	set_entry("npc_usecode_entry", 0x400 + npc_num, true,
 	          npc_num >= 256);
 	// Usually, face = npc_num.
-	set_npc_face(npc_num, 0);
+	{   GtkWidget *widget = get_widget("npc_face_frame");
```
marzojr, line 405: Line break after the {

mapedit/npcedit.cc
```diff
@@ -545,8 +487,15 @@ int ExultStudio::init_npc_window(
 	set_entry("npc_num_entry", npc_num, true, false);
 	set_entry("npc_ident_entry", ident);
 	// Shape/frame.
-	set_npc_shape(shape, 16);
-	set_npc_face(face, 0);
+	set_entry("npc_shape", shape);
+	set_entry("npc_frame", 16);
+	{   GtkWidget *widget = get_widget("npc_face_frame");
```
marzojr, line 492: Line break after the {

mapedit/shapedraw.cc
```diff
@@ -356,3 +357,254 @@ void Shape_draw::start_drag(
...
+				g_object_set_data(G_OBJECT(single->shape), "user_data",
+				    reinterpret_cast<gpointer>(uintptr(shapenum)));
+				char *label = g_strdup_printf("Face #%d", shapenum);
+				gtk_frame_set_label(GTK_FRAME(single->shape), label);
+				g_free(label);
+			}
+		}
+		if (single->frame &&
+		    ((framenum >=  0 &&
+		      framenum <       single->ifile->get_num_frames(shapenum)) ||
```
marzojr, line 548: Weird spacing after the `<`

marzojr (changes requested):

Dragon-Baroque:
Changes posted.
I pushed force the same one-commit branch, please tell me whether you would have preferred a second commit instead.
And for those C++ stylistic defects, please tell me whether you use some checkup tool [ clang-tidy, or whatever ] so that I could perform the checks prior to publish?

marzojr (approved):

marzojr:
At some point I will be reformatting the entire codebase using clang-format, then add a github action that checks if a commit deviates from it. In the mean time, clang-tidy and clang-format using the provided .clang-tidy and .clang-format files will catch most of those (but I still need to do some tweaks to .clang-format).

PR #96, opened 2021-03-20
Simplify tqueue

alagner:
Some code clean-up, mostly moving from bare loops to STL algos/ranged for loops.
I actually wanted to use upper bound in `activate0`, too, but it seems objects are deletign themselves sometimes, which ends in double-free again, so I gave it up, temporarily at least.
@marzojr two questions related to tqueue:
1. What will you say about using a different data structure? As it's supposed to be sorted, multimap with time being the key seems more viable than list. I was also thinking about priority queue, but it does not allow random access.
2. What do you think about including Boost? Some classes like Tqeue could greatly benefit from it, e.g. thanks to its filter_iterator (which currently Tqeue_iteratator replaces). License-wise that should not be a problem, technically not a big issue either, as IIRC when iterator is a header-only lib, question is if an extra dependency is to be tolerated there.

### DIFF
--- a/tqueue.cc
+++ b/tqueue.cc
@@ -92,17 +92,15 @@ int Time_queue::remove(
     Time_sensitive *obj,
     uintptr udata
 ) {
-	if (data.empty())
-		return 0;
-	for (auto it = data.begin();
-	        it != data.end(); ++it) {
-		if (it->handler == obj && it->udata == udata) {
-			obj->queue_cnt--;
-			data.erase(it);
-			return 1;
-		}
+    auto it = std::find_if(data.begin(), data.end(),
+            [&](const auto& el) { return el.handler == obj && el.udata == udata; }
+    );
+    bool found = it != data.end();
+    if (found) {
+		obj->queue_cnt--;
+		data.erase(it);
 	}
-	return 0;         // Not found.
+    return found;
 }
 
 /*
--- a/tqueue.cc
+++ b/tqueue.cc
@@ -166,8 +166,6 @@ void Time_queue::activate0(
 void Time_queue::activate_always(
     uint32 curtime      // Current time.
 ) {
-	if (data.empty())
-		return;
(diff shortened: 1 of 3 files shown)


tqueue.cc
```diff
@@ -82,15 +70,14 @@ bool    operator <(const Queue_entry &q1, const Queue_entry &q2) {
...
-			obj->queue_cnt--;
-			data.erase(it);
-			return 1;
-		}
+	} else {
+		(*toRemove).handler->queue_cnt--;
+		data.erase(toRemove);
+		return 1;
 	}
 	return 0;         // Not found.
```
marzojr, line 82: This is dead code.

tqueue.cc
```diff
@@ -82,15 +70,14 @@ bool    operator <(const Queue_entry &q1, const Queue_entry &q2) {
...
-		if (it->handler == obj) {
-			obj->queue_cnt--;
-			data.erase(it);
-			return 1;
-		}
+	} else {
+		(*toRemove).handler->queue_cnt--;
+		data.erase(toRemove);
+		return 1;
 	}
```
marzojr, line 81: Please avoid using `else` here, as the `if` is just a return. Maybe invert the condition check? See also dead code comment.

tqueue.cc
```diff
@@ -105,17 +92,15 @@ int Time_queue::remove(
...
+    auto it = std::find_if(data.begin(), data.end(),
+            [&](const auto& el) { return el.handler == obj && el.udata == udata; }
+    );
+    bool found = it != data.end();
+    if (found) {
+		obj->queue_cnt--;
+		data.erase(it);
 	}
-	return 0;         // Not found.
+    return found;
```
marzojr, line 103: Indenting with spaces instead of tabs.

marzojr (changes requested):

marzojr (approved):

DominusExult:
This is causing regressions in #143 and #140

PR #98, opened 2021-03-20
RAII replacements

alagner:
Multiple places at the moment, I'm just grepping through and fixing all the low-hanging fruits.

### DIFF
--- a/drag.cc
+++ b/drag.cc
@@ -119,10 +119,7 @@ Dragging_info::Dragging_info(
  *  Delete dragging info.
  */
 
-Dragging_info::~Dragging_info(
-) {
-	delete save;
-}
+Dragging_info::~Dragging_info() = default;
 
 /*
  *  First motion.
@@ -218,7 +215,7 @@ bool Dragging_info::moved(
 		if (!start(x, y))
 			return false;
 	} else if (save)            // Not first time?  Restore beneath.
-		gwin->get_win()->put(save, rect.x, rect.y);
+		gwin->get_win()->put(save.get(), rect.x, rect.y);
 	else
 		gwin->add_dirty(gwin->clip_to_win(rect));
 	gwin->set_painted();
@@ -244,8 +241,11 @@ void Dragging_info::paint(
 ) {
 	if (!rect.w)            // Not moved enough yet?
 		return;
-	if (save)           // Save background.
-		gwin->get_win()->get(save, rect.x, rect.y);
+	if (save) {          // Save background.
+        Image_buffer* tmp;
+		gwin->get_win()->get(tmp, rect.x, rect.y);
+        save.reset(tmp);
+    }
 	if (obj) {
 		if (obj->get_flag(Obj_flags::invisible))
 			obj->paint_invisible(paintx, painty);
--- a/drag.cc
+++ b/drag.cc
@@ -116,12 +116,6 @@ Dragging_info::Dragging_info(
 }
 
 /*
- *  Delete dragging info.
- */
-
-Dragging_info::~Dragging_info() = default;
-
-/*
  *  First motion.
  *
  *  Output: false if failed.
@@ -242,9 +236,9 @@ void Dragging_info::paint(
 	if (!rect.w)            // Not moved enough yet?
 		return;
 	if (save) {          // Save background.
-        Image_buffer* tmp;
+		Image_buffer* tmp;
 		gwin->get_win()->get(tmp, rect.x, rect.y);
-        save.reset(tmp);
+		save.reset(tmp);
     }
 	if (obj) {
 		if (obj->get_flag(Obj_flags::invisible))
--- a/drag.cc
+++ b/drag.cc
@@ -116,6 +116,15 @@ Dragging_info::Dragging_info(
 }
 
 /*
+ *  Delete dragging info.
+ */
+
+Dragging_info::~Dragging_info(
+) {
+	delete save;
+}
+
+/*
  *  First motion.
  *
  *  Output: false if failed.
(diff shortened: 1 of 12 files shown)


marzojr:
Can you rebase based on latest head and force push?

drag.cc
```diff
@@ -244,8 +241,11 @@ void Dragging_info::paint(
 ) {
 	if (!rect.w)            // Not moved enough yet?
 		return;
-	if (save)           // Save background.
-		gwin->get_win()->get(save, rect.x, rect.y);
+	if (save) {          // Save background.
+        Image_buffer* tmp;
+		gwin->get_win()->get(tmp, rect.x, rect.y);
+        save.reset(tmp);
```
marzojr, line 247: Inconsistent indentation: using spaces instead of tabs.

drag.cc
```diff
@@ -244,8 +241,11 @@ void Dragging_info::paint(
...
+    }
```
marzojr, line 248: Missed this one on comment above.

files/utils.cc
```diff
@@ -192,10 +193,9 @@ string get_system_path(const string &path) {
 	if (pos == pos2 && pos3 != string::npos) {
 		int num_chars = GetShortPathName(new_path.c_str(), nullptr, 0);
 		if (num_chars > 0) {
-			char *short_path = new char [num_chars + 1];
-			GetShortPathName(new_path.c_str(), short_path, num_chars + 1);
-			new_path = short_path;
-			delete [] short_path;
+			std::vector<char> short_path(num_chars + 1);
```
marzojr, line 196: Why `vector` instead of `string`?

drag.cc
```diff
@@ -119,10 +119,7 @@ Dragging_info::Dragging_info(
  *  Delete dragging info.
  */
 
-Dragging_info::~Dragging_info(
-) {
-	delete save;
-}
+Dragging_info::~Dragging_info() = default;
```
marzojr, line 122: Might as well remove it and remove the constructor declaration from the class definition.

gamemap.h
```diff
@@ -74,7 +74,7 @@ class Game_map {
 	char *schunk_cache[144];
 	int  schunk_cache_sizes[144];
 	int caching_out;        // >0 in 'cache_out_schunk'.
-	Map_patch_collection *map_patches;
+    std::unique_ptr<Map_patch_collection> map_patches;
```
marzojr, line 77: Indenting with spaces instead of tabs.

objs/mappatch.h
```diff
@@ -92,8 +93,8 @@ class Map_patch_modify : public Map_patch {
 class Map_patch_collection {
 	Map_patch_map patches;
 public:
-	~Map_patch_collection();
-	void add(Map_patch *p);     // Add a patch.
+	~Map_patch_collection() = default;
```
marzojr, line 96: Might as well remove it.

marzojr (changes requested):

drag.cc
alagner: As you prefer. Personally I like to keep it explicit (in the header at least).

files/utils.cc
alagner: Pre-C++11 habit I guess; std::string was not guaranteed to return the underlying array this way back in the day.
Now it is so yes, string should be as good.

drag.cc
alagner: I've re-read that code, I'll revert that change, as it seems to me the `save` field is either modified someplace else or is completetly redundant; I need to double check that, sorry for the hassle.

files/utils.cc
```diff
@@ -844,18 +841,17 @@ void U7copy(
...
 	size_t filesize = in.tellg();
 	in.seekg(0, ios::beg);
 	while (filesize > 0) {      // Copy.
 		size_t toread = bufsize < filesize ? bufsize : filesize;
-		in.read(buf, toread);
-		out.write(buf, toread);
+		in.read(&buf[0], toread);
+		out.write(&buf[0], toread);
 		filesize -= toread;
 	}
```
marzojr, line 853: I was thinking that swapping this by `out << in.rdbuf();` is probably going to be much faster. And it would avoid the undefined behavior from the way the code obtains the file size.

marzojr (changes requested):

marzojr:
>@alagner Revert changes in drag.h/cc 15917fa

Why reverting all changes from drag.h/cc?

files/utils.cc
alagner: OK, I didn't know the `rdbuf` trick, I was actually thinking about using ostreamuf/istreamubf iterators.
Where's the UB, BTW?

marzojr: There are two different sources of UB, both of which come from the C underbelly:

- for a text file, ftell returns a token value that can be used to seek to the same position, but which need not be related to the byte position in the file;
- for binary mode files, ftell does return a byte offset, but fseek to the end is not guaranteed to go one past the last byte of the file.

A couple references:

- [FIO19-C](https://wiki.sei.cmu.edu/confluence/display/c/FIO19-C.+Do+not+use+fseek%28%29+and+ftell%28%29+to+compute+the+size+of+a+regular+file)
- [blog post on Explicit C++](https://web.archive.org/web/20180314195042/http://cpp.indi.frih.net/blog/2014/09/how-to-read-an-entire-file-into-memory-in-cpp/)

And using the `<<` with `rdbuf()` is a LOT faster because it does not behave like formatted output at all; rather, it uses a specialized overload of `<<` which directly uses the underlying stream buffer objects to copy data as fast as possible.

marzojr (approved):

files/utils.cc
alagner: "Love" those caveats. Even more with C, as in C++ it's at least common to expect UB when doing anything ;)

PR #100, opened 2021-03-21
Revert activate_always to prevent freezing time

alagner:
Reverting a breaking change I have introduced while working on time queue.
Turns out, that if one opens a gump during combat,
sometimes the time would completely stop in the game (cheat mode would
keep showing same hour). Relatively easily reproducible with mages
and fighters near bee cave in Yew, especially when mages cast
paralysis when invisible; also applicable to dragons in Destard.

I will invesitgate it more later; so far just taking back the breaking change in full.

### DIFF
--- a/tqueue.cc
+++ b/tqueue.cc
@@ -164,18 +164,23 @@ void Time_queue::activate0(
 void Time_queue::activate_always(
     uint32 curtime      // Current time.
 ) {
-	auto newStart = std::stable_partition(data.begin(), data.end(),
-		[&] (const auto& el) { return !(curtime < el.time); });
-
-	std::for_each(data.begin(), newStart, [curtime](auto& ent) {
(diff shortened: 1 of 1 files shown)


tqueue.cc
```diff
@@ -164,18 +164,23 @@ void Time_queue::activate0(
...
 		if (obj->always) {
 			obj->queue_cnt--;
 			uintptr udata = ent.udata;
+			data.erase(it);
 			obj->handle_event(curtime, udata);
 		}
-	});
-	data.erase(data.begin(), newStart);
+		it = next;
+	}
```
marzojr, line 183: I see what went wrong: in the original code, only the elements that satisfied `obj->always` were removed. This condition was not true in your version.

marzojr (approved):

PR #101, opened 2021-03-27
Prototype : Studio in WIN32 : Use GTK+ 3 OLE2 provided Drag and Drop instead of Windrag

Dragon-Baroque:
This is a proposal for a rework of the Drag and Drop code in Exult and Studio under Windows, as discussed in Issue #71.

### Rationale

In Windows, when the Drag and Drop from Studio to Exult was created, SDL for Exult and GTK+ for Studio only supported receiving WM_DROPFILES messages, and GTK+ supported sending/receiving Drag and Drop messages but only from/to the windows of the current GTK+ application. In other words, there were no way to post a Drag and Drop message from Studio to Exult.

Under Windows, Exult _and_ Studio have their own specialized Drag and Drop code paths to overcome this lack of a working Drag and Drop protocol. This code is in `windrag.h` and `windrag.cc`. Whereas Exult displays a single window acting as an entire Drag and Drop receiver, Studio has many windows and many Drag and Drop sender / receiver widgets in them. As a result, the Exult receiver part as rather simple, the Studio part is complex, and very few GtkDrawingAreas of Studio can act as Drag and Drop receivers, many less than under Linux or MacOS. This is compounded today with the new Shapes and Paperdoll GtkDrawingAreas of Studio, all of which can act as Drag and Drop receivers under Linux and MacOS, but none under Windows.

GTK+ 3 provides an alternate full power cross application OLE2 Drag and Drop, although it is considered «Experimental» and is enabled with an environment variable `GDK_WIN32_USE_EXPERIMENTAL_OLE2_DND`. This Drag and Drop code becomes default in GTK 4. _I have not been able to catch defects in this code_.

### Proposal

- Studio exploits fully the OLE2 Drag and Drop of GTK+ 3, sending Drags to itself and to Studio, receiving Drops from itself.
- Exult keeps a proprietary code for receiving OLE2 Drops from Studio, this is a much simplified and slightly altered `windrag.cc`.
- There are only little signature changes in `exult.cc`, and no other change in Exult.
- All _WIN32 special paths for Drag and Drop in Studio are removed, including references to `windrag.h`.
- Studio in `studio.cc` enables the OLE2 Drag and Drop of GTK+ 3.

### Status

- Exult and Studio compile and run under Linux without change
- Exult and Studio compile and run under Windows, with Drag and Drop from Studio to Studio enabled in all GtkDrawingAreas, with Drag and Drop from Studio to Exult working with the DragOver messages correctly interpreted : Under Windows only, Exult draws the green grid to help a precise positioning of the Dropped object.

Since the OLE2 Drag and Drop is considered experimental in GTK+ 3, I would like this proposal to be strongly tested.

Dragon Baroque, March 27, 2021

### DIFF
--- a/mapedit/studio.h
+++ b/mapedit/studio.h
@@ -92,13 +92,9 @@ using Msg_callback = void (*)(Exult_server::Msg_type id,
                               const unsigned char *data, int datalen, void *client);
 
 #ifndef _WIN32
-#define C_EXPORT extern "C"
+#	define C_EXPORT extern "C"
 #else
-#ifndef WIN32_LEAN_AND_MEAN
-#define WIN32_LEAN_AND_MEAN
-#endif
-#include <windows.h>
-#define C_EXPORT extern "C" __declspec(dllexport)
+#	define C_EXPORT extern "C" __declspec(dllexport)
 #endif
 
 class ExultStudio {
--- a/mapedit/studio.h
+++ b/mapedit/studio.h
@@ -98,7 +98,6 @@ using Msg_callback = void (*)(Exult_server::Msg_type id,
 #define WIN32_LEAN_AND_MEAN
 #endif
 #include <windows.h>
-#include "windrag.h"
 #define C_EXPORT extern "C" __declspec(dllexport)
 #endif
 
@@ -185,19 +184,6 @@ private:
 	int curr_game;  // Which game is loaded
 	int curr_mod;   // Which mod is loaded, or -1 for none
 	std::string game_encoding;  // Character set for current game/mod.
-	// For Win32 DND
-#ifdef _WIN32
-	HWND            egghwnd;
-	Windnd          *eggdnd;
-	HWND            npchwnd;
-	Windnd          *npcdnd;
-	HWND            objhwnd;
-	Windnd          *objdnd;
-	HWND            conthwnd;
-	Windnd          *contdnd;
-	HWND            shphwnd;
-	Windnd          *shpdnd;
-#endif
 	// Server data.
 	int         server_socket;
 	gint            server_input_tag;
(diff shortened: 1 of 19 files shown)


mapedit/studio.h
```diff
@@ -98,7 +98,6 @@ using Msg_callback = void (*)(Exult_server::Msg_type id,
 #define WIN32_LEAN_AND_MEAN
 #endif
 #include <windows.h>
-#include "windrag.h"
 #define C_EXPORT extern "C" __declspec(dllexport)
 #endif
```
marzojr, line 102: Is any of this still needed?

marzojr: (and this shuold be from lines 94-102, but gh would not let me select above line 98)

windrag.cc
```diff
@@ -14,34 +14,13 @@
 #include "u7drag.h"
 #include "ignore_unused_variable_warning.h"
 
-static UINT CF_EXULT = RegisterClipboardFormat("ExultData");
+static UINT CF_EXULT = RegisterClipboardFormat(U7_TARGET_GENERIC_NAME_X11);
```
marzojr, line 17: Why "_X11"?

Dragon-Baroque:
Added a second commit to satisfy your objections.

Comments : 

- While windows.h is no longer needed in studio.h, C_EXPORT still is.
- I renamed the #define of the Generic targets to more purposeful names.

I handled your `studio.h` objection as meaning that I should look at the remaining `_WIN32` occurrences in Studio.
They remain only now in `studio.cc` [ Windows only `portable` option, and Setup/Termination stuff ], and one in `shapelst.cc` to invoke the image editor onto the current shape [ Should it use the execbox instead ? ].
Then I discovered something surprising : `shapelst.cc` does not include explicitly `windows.h`, although it does require it ( `PROCESS_INFORMATION`, ... ). It happens that `windows.h` is included by each `rect.h` and `schedule.h` to rename safely the Exult class `Rectangle` to `RECTX` - because `Rectangle` collides with a C API function in Windows `wingdi.h`. Should not the `Rectangle` class of Exult be renamed to `ExultRect` or `ExultRectangle` instead to avoid this ?

marzojr (approved):

marzojr:
>Should not the Rectangle class of Exult be renamed to ExultRect or ExultRectangle instead to avoid this ?

Probably, yes. But TileRect would probably be better.

PR #120, opened 2021-11-04
Update exultmsg_de.txt

ungesundes-halbwissen:
changed some expressions, 
changed Wächter to Guardian, 
changed formal addressing to informal addressing
added all text required for Serpent Isle

### DIFF
--- a/data/exultmsg_de.txt
+++ b/data/exultmsg_de.txt
@@ -7,11 +7,11 @@
 0x504:Wer hat meinen Stuhl bewegt??
 0x505:Du siehst aus, als wrde es dir gut gehn.
 0x506:Alles klar?
-0x507:Bereit fr das Dessert?
+0x507:Bereit fr den Nachtisch?
 0x508:Guten Appetit!
 0x509:Spezialitt des Hauses!
-0x50a:Avatar!  Benimm dich!
-0x50b:Hast Du bemerkt, da schon jemand im Bett liegt?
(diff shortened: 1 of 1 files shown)


DominusExult:
I'm not sure I can accept that.
(Except for the message LB on Twitter but we need the @ in Exult first which @marzojr wanted to add some time ago).
The exultmsg_de.txt is for the official German release and AFAIR I used the correct lines for that.
 However we should probably add the functionality that Exult looks in the static folder for the exultmsg.txt as well for complete translations like yours.

ungesundes-halbwissen:
Yeah, the old file was the official translation of 1992. What do you mean: the "Guardian" modification? The new spelling rules? Or the informal stuff? The rest should be unique (either Exult-only or Serpent Isle stuff)

But as I said before: Do as you deem appropriate and reject anything I do at will yourself - no hard feelings at any account :-)
It would be nice to have the file ready for Serpent Isle though

DominusExult:
Not sure whether my full response shows after the initial sentence.
The exultmsg text files should be only reflecting the official releases and nit the fan translations (regardless of how much I appreciate it).
With the only exception of pointing at Twitter for reporting thy feat! (And for that we still need to implement the winning screen)
So for your translation you have to work with a patch folder  and your exultmsg being in that.
And in your SI Fixes mod translation you need to place the file again as Exult will ignore the patch folder when you start a mod.

ungesundes-halbwissen:
Ok 👍 :-)

Issue #121, opened 2021-11-21
Exult Studio does not save "Frame Names"


DominusExult:
With the switch to GTK+3 we seem to have lost the ability to save "Frame Names" of shapes.
-double click a shape
- select "Frame Names"
- go to "Frame Names" tab
- add names for frames
- save
- restart Exult/Exult Studio -> Frame Names have not been saved. 

At first glance it seems as if shape_info.txt is not being saved to the patch folder.
@Dragon-Baroque and @marzojr any ideas what goes wrong?

Dragon-Baroque:
The problem occurs much sooner : 
Open the Shape Editor on any Shape without Frame Names, Check "Frame Names" in the main panel of the Shape Editor, Switch to the Frame Names tab, Add names, click on Add, click on OK.
Reopen the Shape Editor on the same Shape, the Frame Names are lost.

This is the same for Content Rules, Frame Flags, Frame HPs, Frame Usecode, Object Paperdoll, Brightness and Warmth.

The problem comes from the transfer back from the GTK treestores into the `Shape_info` vector objects. The Vector of `Frame_name_info` is properly populated by the `Update_framenames` function, but is immediately erased by the `clean_invalid_name_info` in Exult_Studio::save_shape_notebook [ mapedit/shapeedit.cc ]. 
All the `clean_invalid_..` functions run `clean_vector`, but the `clean_vector` [ shapes/data_utils.h ] erase condition is wrong.
It is written as 
 `return elem.is_invalid() || !elem.have_static();`
It should be written instead as
`return elem.is_invalid() && !elem.have_static();`
 by comparison with the code of `clean_vector` prior to commit 9e1a5d356.

I shall provide a Pull Request as soon as I complete retrieving my lost GitHub credentials.

Sincerely
Dragon-Baroque

marzojr:
Can anyone confirm if it is working now?

DominusExult:
Finally just tested it and it works. One might argue that setting "all qualities" as the default could be a good idea.

PR #123, opened 2021-11-24
Studio clean_vector

Dragon-Baroque:
For lost changes in Shape Editor :
On a Shape object without Frame Names, create Frame Names with the Shape Editor, Add, OK, the Frame Names are lost on the Shape Editor exit.
On a Shape object created by a patch, with Frame Names, or on a Shape Object given Frame Names by a patch, alter the Frame Names with the Shape Editor, Add, OK, the Shape loses all its Frame Names at Shape Editor exit.

Same with Frame Flags, Frame HPs, Frame Usecode, Object Paperdoll, Brightness and Warmth.

Reason : the clean_vector function called by all clean_invalid_.. functions has its removal test wrong.

Should fix Issue #121 

Sincerely
Dragon-Baroque

### DIFF
--- a/shapes/data_utils.h
+++ b/shapes/data_utils.h
@@ -88,7 +88,7 @@ template <typename T>
 void clean_vector(std::vector<T> &vec) {
 	vec.erase(std::remove_if(vec.begin(), vec.end(),
 		[](T& elem) {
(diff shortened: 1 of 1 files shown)


marzojr (approved):

marzojr:
Nice catch, thanks.

PR #128, opened 2022-01-10
added key needed to close, fixed Warning centering

Dreded:
Due to ESC key falling out of favour many moons ago I never thought to hit it to close the window, the window manager Close button did not work so it seemed when viewing this warning the only way to close the window was using the task manager/Force close.

Also fixed the centering, not sure why someone manually centered it ages ago(why add 20 to the first line after already setting the first line height just above)
warning

### DIFF
--- a/exultmenu.cc
+++ b/exultmenu.cc
@@ -367,25 +367,29 @@ BaseGameInfo *ExultMenu::run() {
 	navfonton = fontManager.get_font("HOT_NAV_FONT");
 
 	if (!gamemanager->get_game_count()) {
-		int topy = centery - 25;
+		int center_offset = 40; //this is the last line /2 to put the entire message centered on screen
+		int topy = centery - center_offset; 
 		font->center_text(gwin->get_win()->get_ib8(),
-		                  centerx, topy + 20, "WARNING");
+		                  centerx, topy, "WARNING");
 		font->center_text(gwin->get_win()->get_ib8(),
-		                  centerx, topy + 40, "Could not find the static data for either");
+		                  centerx, topy + 20, "Could not find the static data for either");
 		font->center_text(gwin->get_win()->get_ib8(),
-		                  centerx, topy + 50, R"("The Black Gate" or "Serpent Isle".)");
+		                  centerx, topy + 30, R"("The Black Gate" or "Serpent Isle".)");
 #ifndef __IPHONEOS__
 		const char games_missing_msg[] = "Please edit the configuration file";
 #else
 		const char games_missing_msg[] = "Please add the games in iTunes File Sharing";
 #endif
 		font->center_text(gwin->get_win()->get_ib8(),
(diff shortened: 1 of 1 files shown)


DominusExult:
Thanks! You are right that it needs a hint these days :)
I think I'd prefer "Press ESC to exit!" but maybe I'm too accustomed to exit a program instead of closing it :)

Dreded:
ok, I am indifferent as to whether close or exit makes more sense so I changed it... I think Close is more modern exit is old school... exult being old school.. we shall Exit!

Also updated the entire way the message is constructed as it was bothering me than any manual math needed to be done to center it if it ever changed(yes because its gone unchanged for 20yrs and this seems likely to happen) 

Also altered the IPHONEOS messages/delay behavior from #ifndef to #ifdef because I don't care, and neither should anyone else, what happens if im not on a certain OS... I care what happens to my OS.

Dreded:
> Also altered the IPHONEOS messages/delay behavior from #ifndef to #ifdef because I don't care, and neither should anyone else, what happens if im not on a certain OS... I care what happens on a certain OS.

I get why you did this now as IPHONEOS is the ONLY exception, if you want me to change it back + change mine im happy to do so.

DominusExult:
> > Also altered the IPHONEOS messages/delay behavior from #ifndef to #ifdef because I don't care, and neither should anyone else, what happens if im not on a certain OS... I care what happens on a certain OS.
> 
> I get why you did this now as IPHONEOS is the ONLY exception, if you want me to change it back + change mine im happy to do so.

Nah, that's ok. Might be better this way.

DominusExult:
Hmm, it needs a better wording. Exult exits on the "usual" keys return, space and esc as well as a doubleclick. Hence why we probably didn't bother to mention it (and it also has a timer to quit on its own).

Also exit should be lower case.

Dreded:
the time to exit on its own is 3.5 minutes, nobody is waiting for it.

On my system only ESC works, more importantly the windows manager close button does not work, while I have been using computers for 30+yrs and remember well the DOS days, if a window has an "X" that is how I will want to close it I would not think to hit escape, so adding a key on an ERROR SCREEN does not seem like a bad idea.

The whole reason I decided to add this is because I had to force close the window 5x, which in hindsight makes me feel dumb but that's the experience I had.

If on other systems the SPACE, ENTER and Double click works, that doesn't preclude ESC from working so less words should be fine... not about to say "Press ESC, ENTER, SPACE or double click the window to exit."

DominusExult:
agreed on all counts and merged! Thank you!

Dreded:
I checked on another windows system and SPACE, ENTER does work... also Double Click works on both systems... so apparently my main system has something blocking SPACE and ENTER.

DominusExult:
This prompted me to get rid of some SDL_VERSION_ATLEAST(2,0,13), too

PR #129, opened 2022-01-12
added congratulations screen

### LINKED ISSUE #90: BG: show congratulations screen after destroying the black gate
After the endgame view you'd get the congratulations screen in the original. We don't show that.
I've managed to only show it when the endgame plays in game but not when you view the endgame from the menu via https://github.com/DominusExult/exult/commit/c40c5072862bc543af4e88d5f876a06c4936b3dc

But I'm not managing to get the time elapsed and I don't know how to display it.
First problem I'm having is that I can't make use of `clock->get_day()`

Dreded:
resolves #90

also updates the address to twitter.

2022-01-12 00_37_27-Exult Ultima7 Engine

2022-01-12 00_53_58-Exult Ultima7 Engine

2022-01-12 00_38_29-Exult Ultima7 Engine

### DIFF
--- a/gamemgr/bggame.cc
+++ b/gamemgr/bggame.cc
@@ -1894,7 +1894,7 @@ void BG_Game::end_game(bool success) {
 		Game_clock *clock = gwin->get_clock();
 		int total_time = clock->get_total_hours();
 		// Congratulations screen
-		
+
 		// show only when finishing a game and not when viewed from menu
 		// game starts off with 6 hours as its 6am, should we be removing those 6hours?
 		// perhaps this should be gotten from the save game file creation date
@@ -1915,10 +1915,11 @@ void BG_Game::end_game(bool success) {
 				message = get_text_msg(congrats + i);
 				// if we are on the line with the time played
 				if (i == 2) {
-					//initialize to make sure there is no garbage.
+					// initialize to make sure there is no garbage.
 					char buffer[50] = {0};
-					message = buffer;
-					// since we start at 6am we use hours if below 30 of them(30-6=1day), otherwise we don't display hours unless over a month so offset doesnt matter.
+					message         = buffer;
+					// since we start at 6am we use hours if below 30 of them(30-6=1day), otherwise we don't display hours unless over a month so offset doesnt
+					// matter.
 					if (total_time >= 30) {
 						// this is how you would calculate years but since UltimaVII congrats screen has 13 months per year
 						// and VI in game calendar states 12 months per year
@@ -1926,47 +1927,57 @@ void BG_Game::end_game(bool success) {
 						// 8064 hours = 336 days, 672 hours = 28 days
 						// int year = total_time/8064;
 						// total_time %= 8064;
-						int month = total_time/672;
+						int month = total_time / 672;
 						total_time %= 672;
-						int day = total_time/24;
+						int day = total_time / 24;
 						total_time %= 24;
 						int hour = total_time;
 						// as per comment above
 						// if(year > 0) sprintf(buffer,"%d year(s) , ",year);
 						// message = buffer;
-						if(month == 1) sprintf(buffer,"%s%d month",message,month);
-						else if(month > 1) sprintf(buffer,"%s%d months",message,month);
+						if (month == 1)
+							sprintf(buffer, "%s%d month", message, month);
+						else if (month > 1)
+							sprintf(buffer, "%s%d months", message, month);
 						message = buffer;
 
 						// add amperstand only if month(s) and there is more to display.
-						if(day != 0 || hour != 0 && month > 0) sprintf(buffer,"%s & ",message);
+						if (day != 0 || hour != 0 && month > 0)
+							sprintf(buffer, "%s & ", message);
 						message = buffer;
 
-						if(day == 1) sprintf(buffer,"%s%d day",message,day);
-						else if(day > 1) sprintf(buffer,"%s%d days",message,day);
+						if (day == 1)
+							sprintf(buffer, "%s%d day", message, day);
+						else if (day > 1)
+							sprintf(buffer, "%s%d days", message, day);
 						// if no days, display hours(this would only happen on exactly 1,2,3 etc months)
 						// Here so the player doesnt think we didn't track the hours/days.
 						// so 112 days at 2am would display "4 months & 2 hours", 113 days at 2am would display "4 months & 1 day"
-						else if(day == 0 && hour == 1) sprintf(buffer,"%s%d hour",message,hour);
-						else if(day == 0 && hour > 1) sprintf(buffer,"%s%d hours",message,hour);
-						
+						else if (day == 0 && hour == 1)
+							sprintf(buffer, "%s%d hour", message, hour);
+						else if (day == 0 && hour > 1)
+							sprintf(buffer, "%s%d hours", message, hour);
+
 						// in the remote chance a player finishes on exactly 0 hours, 0 days and X month(s)
-						if(day == 0 && hour == 0) sprintf(buffer,"%s & 0 days",message);
+						if (day == 0 && hour == 0)
+							sprintf(buffer, "%s & 0 days", message);
 						message = buffer;
 					} else {
 						// if only displaying hours remove the initial 6
 						total_time -= 6;
-						if(total_time == 1) sprintf(buffer,"only %d hour.",total_time);
-						else sprintf(buffer,"only %d hours.",total_time);
+						if (total_time == 1)
+							sprintf(buffer, "only %d hour.", total_time);
+						else
+							sprintf(buffer, "only %d hours.", total_time);
 						message = buffer;
 					}
 				}
-				//Update the mailing address(great idea DominusExult)
+				// Update the mailing address(great idea DominusExult)
 				if (i == 4) {
-					//fun fact the @ symbol was used to represent quotes :p
+					// fun fact the @ symbol was used to represent quotes :p
 					message = "at RichardGarriott on Twitter";
 				}
-				normal->draw_text(ibuf, centerx - normal->get_text_width(message) / 2, starty + normal->get_text_height()*i, message);
+				normal->draw_text(ibuf, centerx - normal->get_text_width(message) / 2, starty + normal->get_text_height() * i, message);
 			}
 
 			// Fade in for 1 sec (50 cycles)
--- a/gamemgr/bggame.cc
+++ b/gamemgr/bggame.cc
@@ -1949,6 +1949,9 @@ void BG_Game::end_game(bool success) {
 						// so 112 days at 2am would display "4 month's & 2 hour's", 113 days at 2am would display "4 month's & 1 day"
 						else if(day == 0 && hour > 1) sprintf(buffer,"%s%d hour's",message,hour);
 						else if(day == 0 && hour == 1) sprintf(buffer,"%s%d hour",message,hour);
+						
+						// in the remote chance a player finishes on exactly 0 hours, 0 days and X month(s)
+						if(day == 0 && hour == 0) sprintf(buffer,"%s & 0 day's",message);
 						message = buffer;
 					} else {
 						// if only displaying hours remove the initial 6
--- a/gamemgr/bggame.cc
+++ b/gamemgr/bggame.cc
@@ -1918,23 +1918,19 @@ void BG_Game::end_game(bool success) {
 				if (i == 2) {
 					//initialize to make sure there is no garbage.
 					char buffer[50] = {0};
-					if (total_time >= 24) {
-						int year = total_time/8760;
-						total_time %= 8760;
-						int month = total_time/672;
-						total_time %= 672;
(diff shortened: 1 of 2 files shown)


DominusExult:
Looking good so far.
Some thoughts though:
- Could you keep the comments on point without too much chatter or smileys?
- Also please nix the suspiciously  line? It doesn't quite fit grammatically and while funny, it really doesn't fit
- I think instead of fixing the probable 13 month/year bug, let's just not calculate years and keep it to months
- I'll check the original whether it subtracts the starting six hours (the game begins at six am)

@marzojr what are your thoughts? I think you might already have coded some parts of it.

Dreded:
> Could you keep the comments on point without too much chatter or smileys?
can do...

> Also please nix the suspiciously line? It doesn't quite fit grammatically and while funny, it really doesn't fit
done, although https://github.com/exult/exult/pull/129/commits/fb167b1eb537cc88de77422175382c9502adf3f7 did fix the grammatic errors, personally I figured a little humor in  a message nobody should ever see was on point with 1992 design philosophy.

> I think instead of fixing the probable 13 month/year bug, let's just not calculate years and keep it to months
sure, that was my original thought too and I just forgot while coding it.

>I'll check the original whether it subtracts the starting six hours (the game begins at six am)
thanks, was going to try and do this later today but I don't even have DOSBox(yet)

DominusExult:
Thanks for your fixes.

Just tested with DOSBox. If you succeed on the first day, the line is left blank (as they only began with days, not hours).

If I succeed at day 2 at 5am, it says I needed 1 day, which makes sense since it took you a day in the loose use of a day (not 24 hours).

So we maybe should have a special case `if (total_time<=30)`. In this case only show " in (total_time - 6) hours"? But after that special case we don't need to care about the exact hours anymore.
That fixes this comment `// game starts off with 6 hours as its 6am, should we be removing those 6hours?`

And best strike the comment  		
`// perhaps this should be gotten from the save game file creation date` - we don't want that, the creation date of a save file could be years ago, it's just about the game time.

Could you also pretty it up so it checks for amount of 1 so we don't write (s) (month(s), day(s), hour(s))?
And if you were to go over board you could also do checks to leave out days or hours when they are 0 but setting the "&" correctly. 
"2 months & 3 hours" "2 months & 3 days"
OTOH I expect that to look pretty messy in code.

Also as total_hours is calculated by `return day * 24 + hour` you could calculate the months just by get_day.

Dreded:
figured out the 13 months issue... its rather simple not a bug at all in the way I was thinking... but I think its still a bug...
365/28 = 13
so the real question is should the year have 336 days(12 months) or 365 days(13 months), for the funnies I am going to pose this question to Richard Garriot... maybe he has a final answer as its been bugging him for years.

Dreded:
I need to stop posting comments when I don't have the time to double check things... so after reading.
https://wiki.ultimacodex.com/wiki/Britannian_calendar
and #90 again I have realized the year has 336 days.

The problem comes in that in OG U7 used 365 days when doing the calculation on the congrats screen, can we double check this?

DominusExult:
*364* days, 13x28=364 :)
Anyway, we should just ignore years and not print them. Even asking Richard Garriott himself, after nearly 30 years can he really say?

DominusExult:
I propose to only subtract them for the first day and afterwards it is hazy enough with only displaying the months& days it took.

Dreded:
i agree, and that's what's been done. So this is done on my end.
Be aware of the following comments...
`
// if no days, display hours(this would only happen on exactly 1,2,3 etc months)
// Here so the player doesnt think we didn't track the hours/days.
// so 112 days at 2am would display "4 month's & 2 hour's", 113 days at 2am would display "4 month's & 1 day"
`

Dreded:
with the above comments in mind should I add
```
else if(day == 0 && hour == 0) sprintf(buffer,"%s 0 day's",message);
```
so that in the unlikely even a player finishes on exactly 1,2,3 etc months with 0 hours they once again don't need to wonder if only month's where tracked.

Dreded:
thirteenmonths
Just found this... thought it was ironic

DominusExult:
he he, that's a great find at the perfect time.

Just tried the patch and it has two issues left, I think:
- the ampersand is displayed when there is 0 month so needs another case
`if(month!=0 && (day != 0 || hour != 0)) sprintf(buffer,"%s & ",message);`
- it's months, days, hours, not month's, day's, hour's

Thanks for bearing with me! :)

Dreded:
> the ampersand is displayed when there is 0 month so needs another case
woops, I tested under 1 day and over a month.. this is why automated testing is a thing :)

> it's months, days, hours, not month's, day's, hour's
I disagree.. we own the time in this case so apostrophe is correct

- There are many hours in a day.
- I am giving you one day's notice to get this done!

DominusExult:
Uhm, no, it's about the number of months, days and hours that have passed and not about them owning something.
Otherwise you'd need it for 1 month, 1 day, 1 hour as well.
"You completed the game in 1 month's, 1 day's & 1 hour's!"
<- that's obviously wrong.
If the sentence were different it would be ok but were still different than how you used it:
"You completed the game in 1 mobth's time!" or
"You completed the game in 2 months' time!" (It's still 2 months and they own the time, so "months'").

Dreded:
you are right, the reason I gave is wrong, the apostrophe is correct. I honestly do not care enough to stomp my feet, so whatever your next response is. Is the way it will be done.

>https://incorporated.zone/days-or-days/
>Should you use “days” or “day’s” in a sentence?
The word “days” is here the plural form of the word “day”.
The word “day’s” is the possessive form of the word “day”.
You’ll use the word day without an apostrophe when you refer to the word ‘day’ in plural like “there are many days in a year”.
You’ll use the apostrophe followed by the “s” when you are referring to one measure of time like “a day’s notice”.

DominusExult:
It's still not correct.
Read the link you just posted.
First the correct one, if you were right, would be "months' ".

But if you read further:

"Indicating 30 days’ notice is the equivalent of saying “notice of 30 days”."

In our case there is nothing like "notice" or "time" that day is possessing.

Again from your link
"The word “days” is here the plural form of the word “day”.

The word “day’s” is the possessive form of the word “day”."

What does day possess in
"You completed the game in 1 day!"?

gamemgr/bggame.cc
```diff
@@ -1889,23 +1890,82 @@ void BG_Game::end_game(bool success) {
...
+						int month = total_time/672;
+						total_time %= 672;
+						int day = total_time/24;
+						total_time %= 24;
+						int hour = total_time;
+						// as per comment above
+						// if(year > 0) sprintf(buffer,"%d year(s) , ",year);
+						// message = buffer;
+						if(month == 1) sprintf(buffer,"%s%d month",message,month);
+						else if(month > 1) sprintf(buffer,"%s%d month's",message,month);
```
marzojr, line 1938: Please add a space between `if` and `(`., and break the `if` and `else` statements into their own lines.

gamemgr/bggame.cc
```diff
@@ -1889,23 +1890,82 @@ void BG_Game::end_game(bool success) {
...
+						message = buffer;
+
+						// add amperstand only if month(s) and there is more to display.
+						if(day != 0 || hour != 0 && month > 0) sprintf(buffer,"%s & ",message);
+						message = buffer;
+
+						if(day == 1) sprintf(buffer,"%s%d day",message,day);
+						else if(day > 1) sprintf(buffer,"%s%d day's",message,day);
```
marzojr, line 1946: Please add a space between `if` and `(`., and break the `if` and `else` statements into their own lines.

gamemgr/bggame.cc
```diff
@@ -1889,23 +1890,82 @@ void BG_Game::end_game(bool success) {
...
+						int hour = total_time;
+						// as per comment above
+						// if(year > 0) sprintf(buffer,"%d year(s) , ",year);
+						// message = buffer;
+						if(month == 1) sprintf(buffer,"%s%d month",message,month);
+						else if(month > 1) sprintf(buffer,"%s%d month's",message,month);
+						message = buffer;
+
+						// add amperstand only if month(s) and there is more to display.
+						if(day != 0 || hour != 0 && month > 0) sprintf(buffer,"%s & ",message);
```
marzojr, line 1942: Please add a space between `if` and `(`., and break the `if` statement into its own line.

gamemgr/bggame.cc
```diff
@@ -1889,23 +1890,82 @@ void BG_Game::end_game(bool success) {
...
+						message = buffer;
+
+						if(day == 1) sprintf(buffer,"%s%d day",message,day);
+						else if(day > 1) sprintf(buffer,"%s%d day's",message,day);
+						// if no days, display hours(this would only happen on exactly 1,2,3 etc months)
+						// Here so the player doesnt think we didn't track the hours/days.
+						// so 112 days at 2am would display "4 month's & 2 hour's", 113 days at 2am would display "4 month's & 1 day"
+						else if(day == 0 && hour == 1) sprintf(buffer,"%s%d hour",message,hour);
+						else if(day == 0 && hour > 1) sprintf(buffer,"%s%d hour's",message,hour);
```
marzojr, line 1951: Please add a space between `if` and `(`., and break the `if` and `else` statements into their own lines.

gamemgr/bggame.cc
```diff
@@ -1889,23 +1890,82 @@ void BG_Game::end_game(bool success) {
...
+						
+						// in the remote chance a player finishes on exactly 0 hours, 0 days and X month(s)
+						if(day == 0 && hour == 0) sprintf(buffer,"%s & 0 day's",message);
+						message = buffer;
+					} else {
+						// if only displaying hours remove the initial 6
+						total_time -= 6;
+						if(total_time == 1) sprintf(buffer,"only %d hour.",total_time);
+						else sprintf(buffer,"only %d hour's.",total_time);
```
marzojr, line 1960: Please add a space between `if` and `(`., and break the `if` and `else` statements into their own lines.

gamemgr/bggame.cc
```diff
@@ -1889,23 +1890,82 @@ void BG_Game::end_game(bool success) {
...
+						if(day == 1) sprintf(buffer,"%s%d day",message,day);
+						else if(day > 1) sprintf(buffer,"%s%d day's",message,day);
+						// if no days, display hours(this would only happen on exactly 1,2,3 etc months)
+						// Here so the player doesnt think we didn't track the hours/days.
+						// so 112 days at 2am would display "4 month's & 2 hour's", 113 days at 2am would display "4 month's & 1 day"
+						else if(day == 0 && hour == 1) sprintf(buffer,"%s%d hour",message,hour);
+						else if(day == 0 && hour > 1) sprintf(buffer,"%s%d hour's",message,hour);
+						
+						// in the remote chance a player finishes on exactly 0 hours, 0 days and X month(s)
+						if(day == 0 && hour == 0) sprintf(buffer,"%s & 0 day's",message);
```
marzojr, line 1954: Please add a space between `if` and `(`., and break the `if` statement into its own line.

gamemgr/bggame.cc
marzojr, line 1938: `month's` is possessive, not plural; please remove the apostrophe.

gamemgr/bggame.cc
marzojr, line 1946: `day's` is possessive, not plural; please remove the apostrophe.

gamemgr/bggame.cc
marzojr, line 1951: `month's` and `hour's` are possessives, not plurals; please remove the apostrophes.

gamemgr/bggame.cc
marzojr, line 1960: `hour's` is possessive, not plural; please remove the apostrophes.

marzojr (changes requested):

marzojr:
> you are right, the reason I gave is wrong, the apostrophe is correct. I honestly do not care enough to stomp my feet, so whatever your next response is.

The reason you gave was wrong, and the apostrophe in these cases is 100% wrong. The source you gave at the end says exactly that the apostrophe is not to be used there.

gamemgr/bggame.cc
Dreded: for all of this you could have just said, apply clang-format.. I didn't do this previously as it formats lines to 80 columns, not to mention the entire code base has not been formatted... so can we agree to add 

ColumnLimit:     '160'

or do we actually want the default 80 wide columns?

For the time being I would obviously just apply it to my added code(selection)

DominusExult:
Thanks a lot Dreded!

marzojr:
> for all of this you could have just said, apply clang-format.. I didn't do this previously as it formats lines to 80 columns, not to mention the entire code base has not been formatted.

The reason I did not say 'apply clang-format' was because the code base has not been formatted. I have been meaning to do that, but first, I want to make sure that there are no more statements left that are `#ifdef`ed in or out of existence (partially or entirely) on some platforms, or compilation options, as that makes formatting platform-dependent.

PR #168, opened 2022-04-03
Fix out-of-tree autogen

ceckak:
The existing `autogen.sh` assumes it is being run from the source directory.  This causes problems if the build is being managed by a tool which is running from another directory.  This patch modifies the script to change to the source directory before rerunning autotools.

### DIFF
--- a/autogen.sh
+++ b/autogen.sh
@@ -1,5 +1,9 @@
 #!/bin/sh
 
+# Move to autogen.sh's directory
+srcdir=`dirname $0`
+test -n "$srcdir" && cd "$srcdir"
+
 DIE=0
 
 # Check for availability
--- a/autogen.sh
+++ b/autogen.sh
@@ -1,7 +1,7 @@
 #!/bin/sh
 
 # Move to autogen.sh's directory
-srcdir=`dirname $0`
+srcdir=$(dirname $0)
 test -n "$srcdir" && cd "$srcdir"
 
 DIE=0


ceckak:
First of a series of pull requests to merge in the android port (https://github.com/ceckak/exult/tree/android-port)

DominusExult:
Merged! thank you!
out-of-tree building does not quite work yet (on macOS at least) because files/zip/zip.cc doesn't find the headers folder when build out-of-tree. Need to investigate :)

Dragon-Baroque:
Should we remove `autogen.sh` and replace it by `autoreconf -i -v` ?
This is what I do in my git branches. `/usr/bin/autoreconf` is provided by the `autoconf` package.

DominusExult:
might need some work, too, to properly function. For example on macOS libtool/libtoolize are often built as glibtool/glibtoolize to not clash with the libtool provided by Apple which is something different :)

ceckak:
Out-of-tree building should get fully resolved with upcoming pull requests.  I've been successfully doing builds for MacOS and android  from out of tree.  This one was just an isolated fix specific to autogen.  I'll circle back on the questions about autoreconf and libtool - don't have a lot of context on those.

DominusExult:
Oops, sorry, I jumped the gun and have already comitted a fix for the files/zip makefile.am (using the solution done in the other makefile.am files instead of yours that I only thought to look for after I comitted my fix). Sorry for the hassle you will have to go through :(

As for autoreconf/libtool, you don't need to worry about that (unless you want to). I think @Dragon-Baroque didn't mean you but rather me and the other devs :)

ceckak:
No worries - I'll rebase and try to get some more pull requests posted tonight.  I'm trying to figure out the best way to do these with github - haven't worked through this process before.  It sounds like a single pull request for the whole series would force you guys to accept/reject it wholesale.  It seems easier to submit them individually so we can make gradual progress without having review the whole series in one go.  I'll see if I can get several of them posted in parallel now that I've worked through the process once.

DominusExult:
Yes, that's probably the best way forward. The hardest part to review (as someone else will have to do it) are the changes before you add the Android only stuff

ceckak:
Posted a series of PRs.  The first one only has one patch.  The later ones are cumulative (don't see a way around that), but the intent is to just review the final patch in each PR, and hopefully they can all merge cleanly if you do them in order.

PR #169, opened 2022-04-03
Remove explicit CXXLINK line

ceckak:
`Makefile.am` currently calls out an explicit `CXXLINK` variable which ends up causing problems with the android build.  This variable should not be necessary as the tools will infer an appropriate link command.  This patch removes the line from the build.


ceckak:
Windows build is failing consistently across this series of PRs due to inability to find `stdlib.h`.  Don't think it's related to any of these patches though.

PR #173, opened 2022-04-03
Detect SDL with SDL_main

ceckak:
On Android, SDL replaces the `main()` symbol with `SDL_main()`, which results in a link error if we try to link an executable.  This is actually a more generic concept in SDL that is turned on by default in android builds.  This patch adds support for detecting when a platform enables `SDL_main()` and adds a wrapper with a real `main()` to call it when building an executable.


DominusExult:
According to the CI it works on linux and macos and I tested on Windows as well. So merged for now

PR #174, opened 2022-04-03
Build Exult as a library

ceckak:
The existing build structure always tries to build exult as an executable.  This ends up causing link errors when building for android.  This patch modifies the build structure to build the entire exult codebase into a library and then link that library into an executable on platforms that support it.  This will enable android builds to opt out of the executable while still building the library which the app will load at runtime.


Dragon-Baroque:
I am doing builds with `--enable-shared` on Linux.

With this commit and that option, Exult is now produced as a small `bin/exult` and a quite big `lib/libexult.so`.
Except that the libexult.so doesn't build because of a missing Freetype library :

In `./Makefile.am` :
replace
```
libexult_la_LIBADD = \
       $(EXULTLIBS) $(PNG_LIBS) \
```
by 
```
libexult_la_LIBADD = \
       $(EXULTLIBS) $(PNG_LIBS) $(FREETYPE2_LIBS) \
```

I don't propose this as a Pull Request because I wonder about Android and MacOS.

ceckak:
I'll have to do some digging to understand why this was working before and failing now, but your proposed change seems likely to be safe for android, etc as FREETYPE2_LIBS will just evaluate to empty if freetype is not being used.

DominusExult:
yes, I might have jumped the gun here.
The lib is a bit problematic, it gets build on macOS as well (it's a static lib with all the linked libraries built into). It could just be ignored but if you were to run "make install" that lib gets installed as well even though it is completely useless :)
So maybe you will need to make building the lib conditional, so it only gets build for the Android target?
The longer way around but perhaps better.

Dragon-Baroque:
Hmm. Perhaps we could just be a little more selective in the `make install` step.

With `--enable-shared` I end up with `lib/libexult.so.0.0.0` proper, and with `libexult.so` and `libexult.so.0` soft links to the proper lib. I find that verbose but this is the normal way. Only we should ensure that we leave `0.0.0`, this is amateurish.

But there is also `lib/libexult.a` and `lib/libexult.la` and those should not be there.

Without `--enable-shared` I still find `lib/libexult.a` and `lib/libexult.la`.

Should the `make install` step for `lib/` have an explicit list of files ?

DominusExult:
Hmm, if you think that's a good way, sure ;)

Dragon-Baroque:
Progress report, sort of :

- One cannot remove `libexult.la`, it is a small text file needed for `make uninstall`,
- I am able not to install `libexult.a` _if I produce_ the `libexult...so..`, just add `--enable-shared --disable-static` to `./configure`. In essence `--enable-shared` asks `libtool` to build and install a `libexult.so`, the shared library, and `--disable-static` prevents the distribution of `libexult.a`, the static library.
- But, if I do not build the shared library [ if I do not use `--enable-shared` ], I cannot remove `libexult.a`. As if `--disable-static` is not taken in that case.
- I do not see any option to restrict the files to be installed. The RPM package distribution format does this by globbing explicitely the files that go into the various binary packages built from a source package.
 
I think that going to Shared Libraries is a path that should be agreed and enforced globally for Exult, all platforms. Personally, for the Linux platform, I would agree. I have the feeling that it would be mandatory for Android, correct me please if wrong. I cannot assess MacOS or MingW.

- On the versioning, `libtool` claims that building `libexult.so.1.8.0` as I tried to do, contravenes the purpose of the `version-info`, even if it is the convention that I find mostly in use in `/usr/lib64`. I have been able to trick `libtool` to produce `libexult.so.1.8.0`, however.
- Instead, `libtool` claims that the release convention should result in `libexult-1.8.0.so` or `libexult-1.8.so`, with the release _before_ the `so` suffix. This convention looks unusual, but I have found it used quite a few times in `/usr/lib64`.
- I could also obtain the combination `libexult-1.so.8.1.2`.
- It boils down to the compatibilities that you want to claim : `exult` makes reference to `libexult-1.so.8` in the latest example. Essentially one claims that shared libraries with different releases on the left of the `so` or different first number on the right are not compatible, whereas shared libraries different only in the second or third number on the right _are_ compatible.

What compatibility in shared libraries do we want to claim ?
If none, then we should use `libexult-1.so.8[.0.0]` or `libexult-1.8.so` or `libexult-1.8.0.so`.
I would recommend for the second if we do not intend to distribute intermediate release like `1.8.1` , `1.8.2` after `1.8`, and the first or the third if we do intend, depending on whether `1.8.1` is or is not compatible with `1.8.0`.

DominusExult:
For a 1.8 release I'm going to back out this change in the new v1.8.x branch. And let this sort out in master.
I think we should keep it open that we can release a bugfix release with x.x.1.
But I also think that maybe we really should look into making this building of the lib limited to Android (if ANDROID), it seems to me as if we add too much trouble for something we only need on Android.

ceckak:
That makes sense to me.  There's not much point/value in distributing exult as a library; it is only needed internally during an Android build.  When I set this up, I was trying to minimize the amount of Android special cases in my changes and figured this would be a harmless change, but I didn't pay that much attention to the install artifacts.

I am curious how the ios app builds.  Does it also use a shared library, or is it able to build/link native code directly into the app?

DominusExult:
I didn't think it would be that disruptive to the other platforms, too :)

iOS uses an xcode project file, so doesn't use the autotools files at all and it builds the app directly without building it as a library.

Issue #176, opened 2022-04-09
Release v1.8 on April 16th (anniversary of Ultima VII)


DominusExult:
I'm preparing a new release in exactly a week on the anniversary of Ultima VII.
Therefore I branched v1.8.x for preparing this. Next step is going to be building these and testing whether they run and the version/copyright dates are really correct on Windows and macOS.

@mbanck, I wonder if you could again test for linux :)

@ceckak agreed to make an Android build, even though his code is not yet completely merged and will only continue to be merged in master - Thanks!

If anyone knows of on an issue that needs to be addressed first or stop this release for any reason, now is the time to speak up! :)

mbanck:
My suggestion would be to tag/release a release candidate in order to test the release procedure and also for packages (e.g. me) to test the tarballs.

DominusExult:
I did a rc1 tag https://github.com/exult/exult/releases/tag/v1.8rc1 - can you test the linux release procedure anyway?

ceckak:
I've spot-checked a rollback of the SDL version for android and confirmed it fixes issues I was seeing with the virtual joystick and music playback.  I'll set up a branch based off this tag in my android fork tonight.

i30817:
BTW, is the gimp plugin working with recent gimp again (ie: not 2.0)?

I have it committed from the my build, which is not ideal.

DominusExult:
Marzo just fixed building the plugin - it was broken for linux/macOS when he converted it to c++. Using gimptool is somehow broken. 
But it works now with current Gimp 2.10.30.

Dragon-Baroque:
> But it works now with current Gimp 2.10.30.

I am afraid I also have some compilation issues on Linux. One of them I reported as a comment to Marzo's commit, and I have two more by now [ trying to compile it on my dev branches ].

Would you agree to a Pull Request ?

DominusExult:
yes, list all that here, I'd say, for discussion (I did NOT understand your comment with placing the header files close to each other) and then you have the power to commit the fixes yourself :)

Dragon-Baroque:
Here are the diffs, with some comments below each one

```
--- a/configure.ac
+++ b/configure.ac
@@ -1060,8 +1060,8 @@ if test x$enable_gimp_plugin = xyes; then
                AC_MSG_RESULT([no, not building GIMP plugin])
        else
                AC_MSG_CHECKING([for GIMP version])
[...]

PR #177, opened 2022-04-10
Fix Gimp Plugin compilation on Linux

Dragon-Baroque:
Commit 1 of 1
    configure.ac
      Handle Gimp Ver.Rel.Mod where Rel >= 10
    mapedit/u7shp.cc
      Move gimp.h and gimpui.h under the warning deactivate frame
      Add 2 new warning deactivates
      Do no enable the GTK .. Deprecated checks

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1000,8 +1000,8 @@ if test x$enable_gimp_plugin = xyes; then
 		AC_MSG_RESULT([no, not building GIMP plugin])
 	else
 		AC_MSG_CHECKING([for GIMP version])
-		gimp_version=`$GIMPTOOL --version | awk 'BEGIN { FS = "."; } { print $1 * 1000 + $2*100+$3;}'`
-		if test "$gimp_version" -ge 2800; then
+		gimp_version=`$GIMPTOOL --version | awk 'BEGIN { FS = "."; } { print $1 * 10000 + $2 * 100 + $3;}'`
+		if test "$gimp_version" -ge 20800; then
 			AC_MSG_RESULT([found >= 2.8.0])
 			AC_SUBST(GIMPTOOL)
 			AM_CONDITIONAL(GIMP_PLUGIN, true)
--- a/mapedit/u7shp.cc
+++ b/mapedit/u7shp.cc
@@ -31,25 +31,19 @@
 #	pragma GCC diagnostic ignored "-Wdeprecated-declarations"
 #	pragma GCC diagnostic ignored "-Wold-style-cast"
 #	if !defined(__llvm__) && !defined(__clang__)
+#		pragma GCC diagnostic ignored "-Wpedantic"
 #		pragma GCC diagnostic ignored "-Wuseless-cast"
 #	else
+#		pragma GCC diagnostic ignored "-Wc99-extensions"
 #		pragma GCC diagnostic ignored "-Wzero-as-null-pointer-constant"
 #	endif
 #endif    // __GNUC__
-#ifdef USE_STRICT_GTK
-#	define GTK_DISABLE_SINGLE_INCLUDES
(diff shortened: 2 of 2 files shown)


Dragon-Baroque:
#176 
Into exult branch master by the Pull Request method

i30817:
Can this be used in gimp, considering you said the window widget that you said this used was deprecated by GTK?

Dragon-Baroque:
> Can this be used in gimp, considering you said the window widget that you said this used was deprecated by GTK?

Short answer : Honestly, I have no idea.

Longer answer : I proposed that patch to make my Linux build of Exult work again, as they create the Gimp plugin.
The `make` step failed before this patch, and that broke the compilation of `exult_studio` too. But the `make install` step always fails on the Gimp plugin because it wants to install it into the protected folder `/usr/lib64/gimp/2.0/plugins/`. I left it at that, not knowing whether it worked or not. By the way Exult Studio does not need it, since it communicates with Gimp via PNG files.

On the one hand, the FileSelector and actually other GTK calls used by the Gimp plugin are flagged deprecated by GTK+ 2, _but are still working_ in GTK+ 2. The deprecated flag is provided by GTK+ 2 as a help to move to GTK+ 3, where these services _do not exist any more_.

On the other hand, I have not installed or used the Gimp plugin even once, and I do not know how to use it nor for what purpose. Is  it to be able to open directly shape files or their container the .VGA file under Gimp ? Same about the Gnome Thumbnailer, by the way. I expected to wait for Gimp 3.0 [ GTK+ 3.0 ] before looking at it. And about GTK 4, I intend to wait until the Drag and Drop works consistently on all platforms. Now it does not. And this holds my migration of Studio to GTK 4 ( done but not publishable ).

Conclusion : I think only you can tell us whether the Gimp plugin is working or whether your patch proposals are still needed.

DominusExult:
I just tested the gimp plugin yesterday and it works with Gimp 2.10.30. 
It's purpose is to open shp files directly in Gimp (frames are displayed as layers).

The Gnome thumbnailer is used to display thimbnails of shp files AFAIK.

marzojr:
The Gimp plugin is for directly opening shape files. You get a prompt for a palette (or can select none for a hard-coded one), and it shows one layer per frame in the shape. It does not handle VGA files directly, though.

The Gnome SHP Thumbnailer is exactly what the name shays -- when Nautilus encounters a shape file, it will call the thumbnailer which will then generate a thumbnail of the first non-blank frame.

Dragon-Baroque:
> It's building now, i had some dependencies missing. Just wondering, you still need libgtk2.0-dev besides libgimp2.0-dev?

I would expect yes :
- Most certainly, if only for the _famous_ FileSelector widget.
- If I try to remove `gtk2-devel` it wants to remove `gimp-devel` too. This is a simple way to detect dependencies.
- `pkg-config --cflags gimpui-2.0` lists all the includes that are part of `pkg-config --cflags gtk+-2.0`. Moreover the `gimpui-2.0.pc` explicitely lists `gtk+-2.0` as required.

Actually we could remove `gtk+-2.0` from the `GIMP_LIBS` and `GIMP_INCLUDES` `$PKGCONFIG` call in `configure.ac`, as it is dragged by `gimpui-2.0` anyway.

Dragon-Baroque:
@i30817, I tried to look at your original patch proposal.

It seems to boil down around versions of `gegl`, I mean 0.3 versus 0.4, true ?

Well on my Linux only `gegl 0.4` is installed, and only `gegl 0.4` is actually downloadable and installable.
Quite possibly because the only Gimp I can download and install is 2.10.30. ( Linux Fedora 35 ).

Thus, is the issue that made you work on your patch still valid ?

PR #179, opened 2022-04-16
Add option to build Exult as a library

ceckak:
The existing build structure always tries to build exult as an executable.  This ends up causing link errors when building for android.  This patch introduces a configure option to build the entire exult codebase into a library instead of an executable.  This will enable android builds to opt out of the executable while still building the library which the app will load at runtime.  This configure option is disabled by default, so builds on other platforms will still generate a monolithic executable like they always have.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -20,14 +20,7 @@ SUBDIRS =  files conf data shapes imagewin pathfinder \
 	gamemgr flic usecode tools audio gumps objs server \
 	mapedit desktop docs $(MODSDIR)
 
-bin_PROGRAMS = exult
-if SDL_MAIN_NEEDED
-exult_SOURCES = sdl_main.cc
-else
-exult_SOURCES =
-endif
-
-exult_SOURCES +=	\
+EXULTSOURCES =	\
 	actions.cc	\
 	actions.h	\
 	actorio.cc	\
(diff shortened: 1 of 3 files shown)


ceckak:
Take two on the exult-as-library PR.  This version makes building as a library optional (disabled by default) and includes the versioning and freetype library fixes suggested by @Dragon-Baroque

ceckak:
Also worth noting that the [SDL_main patch](https://github.com/exult/exult/pull/173) is probably no longer needed now that the android build will only compile the library and will not try to link an executable.  It isn't hurting anything, but could be cleaned out to reduce clutter.

DominusExult:
Could you revise your pr and remove the sdl main as well?
If need be at some later time it still is in the git history but this way we don't carry it around

DominusExult:
And test whether it really is no longer needed by your by Android build ;)

ceckak:
Can do.  Probably won't be able to update it until tonight.

DominusExult:
Just tried the pull request on macOS and it behaves nicely now.
One thing I noticed, in macosx.am, shouldn't the if !BUILD_LIBEXULT not be around `bundle: ${EXE_TARGET}`? This way it prevents all targets that include bundle, not just the one that installs it, or am I misunderstanding something?

ceckak:
> One thing I noticed, in macosx.am, shouldn't the if !BUILD_LIBEXULT not be around bundle: ${EXE_TARGET}? This way it prevents all targets that include bundle, not just the one that installs it, or am I misunderstanding something?

If I do that, `make install` fails because there is "No rule to make target 'bundle', needed by 'install-exec-local'."

If I create a dummy/empty `bundle` rule, the dependency is fixed, but I he install fails during install because `Exult.app` doesn't exist.

Last option would be to conditionally generate a bundle that has the library and no executable, but is that a normal way to distribute Mac libraries?

My goal was to support doing a `make install` that installs the library in the install prefix, but doesn't try to build the executable.

Open to any preferences/suggestions.

DominusExult:
I think we are good, I misunderstood the purpose

ceckak:
> Also worth noting that the https://github.com/exult/exult/pull/173 is probably no longer needed now that the android build will only compile the library and will not try to link an executable. It isn't hurting anything, but could be cleaned out to reduce clutter.

Ok - I take it back.  Tried reverting that commit and the android build fails.  We don't technically need the `sdl_main.cc`, but we do still need the changes that commit introduced in `acinclude.m4`.  Without them, it tries to link an executable to verify that we have a working SDL library, and the link fails.

DominusExult:
So, then I just accepted this PR. If you think you can clean out anything from the SDL Main commit, please do so :)

ceckak:
I've added https://github.com/exult/exult/pull/180 to remove the unneeded `sdl_main.cc`, but it leaves the `acinclude.m4` pieces in-place.  Also posted pull requests for the next series which address the changes to filesystem access.  These ones are the most intrusive into the existing codebase.

PR #181, opened 2022-04-18
Abstract stream types for file utils

ceckak:
The current set of file utilities assume a `std::ifstream`.  Some platforms have nonstandard file storage that would benefit from a more generic abstraction.  This patch changes to using `std::istream` pointers so that subsequent patches can introduce new file access implementations.

Note that this (and the next) patch are not strictly necessary to get Exult working on Android.  The port could make use of standard file paths and avoid widespread changes to the codebase.  However, given that SDL is offering a platform-agnostic abstraction for filesystem access and Exult had a hard dependency on SDL, this seems like an elegant solution, and enables easy access to embedded app resources without ifdeffing the main codebase.  This might allow removing some ifdefs for ios as well.

### DIFF
--- a/audio/midi_drivers/MT32EmuMidiDriver.cpp
+++ b/audio/midi_drivers/MT32EmuMidiDriver.cpp
@@ -94,15 +94,15 @@ int MT32EmuMidiDriver::open() {
 		FileStream part1;
 		FileStream part2;
 		if (openROMFile(part1, "MT32A.BIN", false) && openROMFile(part2, "MT32B.BIN", false)) {
-			std::ofstream out;
-			if (U7open(out, "<SAVEHOME>/data/MT32_CONTROL.ROM", false)) {
+			auto pOut = U7open_out("<SAVEHOME>/data/MT32_CONTROL.ROM", false);
+			if (pOut) {
+				auto& out = *pOut;
 				const Bit8u *data1 = part1.getData();
 				const Bit8u *data2 = part2.getData();
 				for (size_t ii = 0; ii < std::min(part1.getSize(), part2.getSize()); ii++) {
 					out.put(static_cast<char>(data1[ii]));
 					out.put(static_cast<char>(data2[ii]));
 				}
-				out.close();
 				controlROMImage = getROM(controlROMFile, "MT32_CONTROL.ROM");
 			}
 		}
--- a/conf/Configuration.cc
+++ b/conf/Configuration.cc
@@ -216,13 +216,14 @@ bool Configuration::read_abs_config_file(const string &input_filename, const str
 
 	is_file = true; // set to file, even if file not found
 
-	std::ifstream ifile;
+	std::unique_ptr<std::istream> pIfile;
 	try {
-		U7open(ifile, filename.c_str(), true);
+		pIfile = U7open_in(filename.c_str(), true);
 	} catch (exult_exception &) {
 		// configuration file not found
 		return false;
 	}
+	auto& ifile = *pIfile;
 
 	if (ifile.fail())
 		return false;
@@ -235,8 +236,6 @@ bool Configuration::read_abs_config_file(const string &input_filename, const str
 		sbuf += line + "\n";
 		getline(ifile, line);
 	}
-
-	ifile.close();
 
 	CTRACE("Configuration::read_config_file - file read");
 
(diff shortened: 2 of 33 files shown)


Dragon-Baroque:
Do you think possible of using a reference ( to a stream or a more generic type ) instead of a pointer ?

ceckak:
> Do you think possible of using a reference ( to a stream or a more generic type ) instead of a pointer ?

It's awkward with the factory.  Technically you can make a factory that returns a reference, but I think the code would give me nightmares:

```
#include <iostream>

class Foo {
public:
    Foo() {
        std::cout << __PRETTY_FUNCTION__ << std::endl;
    }
    virtual ~Foo() {
        std::cout << __PRETTY_FUNCTION__ << std::endl;
    }
};

class Bar : public Foo {
public:
    Bar() {
        std::cout << __PRETTY_FUNCTION__ << std::endl;
    }
    virtual ~Bar() {
        std::cout << __PRETTY_FUNCTION__ << std::endl;
    }
};

Foo& factory(bool makeBar) {
    return *(makeBar ? new Bar : new Foo);
}

int main(void) {
    std::cout << __PRETTY_FUNCTION__ << ":" << __LINE__ << std::endl;
    Foo& foo = factory(false);
    std::cout << __PRETTY_FUNCTION__ << ":" << __LINE__ << std::endl;
    Foo& bar = factory(true);
    std::cout << __PRETTY_FUNCTION__ << ":" << __LINE__ << std::endl;
    delete &foo;
    std::cout << __PRETTY_FUNCTION__ << ":" << __LINE__ << std::endl;
    delete &bar;
    std::cout << __PRETTY_FUNCTION__ << ":" << __LINE__ << std::endl;
    return 0;
}
```

Another option is to create some kind of wrapper object that can be passed by value and keeps the pointer internally.  You'd have to set it up as a facade that has all the `std::istream` member functions that we use to make it work correctly.  I opted for the `std::unique_ptr<std::istream>` so that it is clear what we are working with.

Dragon-Baroque:
I must admit that I do not see the point of your code example nor therefore why it would cause you nighmares.

marzojr:
His nightmares probably come from lines like this:

```c++
delete &foo;
```

Not only do you need to delete the address of a reference, but you also need to do manual memory management.

A compromise, which would also reduce the size of the diff, is to do something like this instead: keep returning a (smart) pointer, but add a variable for it, and initialize a reference to it from the dereference. For example, `Configuration::read_abs_config_file` would be something like:

```diff
 
 	is_file = true; // set to file, even if file not found
 
-	std::ifstream ifile;
+	std::unique_ptr<std::istream> ifileptr;;
 	try {
-		U7open(ifile, filename.c_str(), true);
+		ifile = U7open_in(filename.c_str(), true);
 	} catch (exult_exception &) {
 		// configuration file not found
 		return false;
 	}
 
+	auto& ifile = *ifileptr;
 	if (ifile.fail())
		return false;
 
```

marzojr:
These pull requests are hopelessly entangled together:
#181 
#182 
#183 
#184 

What is the point of making them separate pull requests if they cannot be merged independently?

marzojr:
These pull requests are hopelessly entangled together:
#181 
#182 
#183 
#184 

What is the point of making them separate pull requests if they cannot be merged independently?

For future reference: when preparing such a batch of pull requests that are meant to be merged independently, please create a separate branch for each pull request and propose only the changes from base. That allows updating one pull request without updating them all.

ceckak:
@Dragon-Baroque 
> I must admit that I do not see the point of your code example nor therefore why it would cause you nighmares.

The example was just an experiment to demonstrate to myself that it is indeed possible to have a factory that returns references.  The nightmares are about using a raw C++ reference to manage a dynamically allocated object, and the risk of accidentally leaking memory.  This may not have been your intention; I was just exploring the option to see if it would work.

@marzojr 
> A compromise, which would also reduce the size of the diff, is to do something like this instead: keep returning a (smart) pointer, but add a variable for it, and initialize a reference to it from the dereference.

Happy to make this change if there is consensus that this is the preferred style.

ceckak:
> These pull requests are hopelessly entangled together: #181 #182 #183 #184
> 
> What is the point of making them separate pull requests if they cannot be merged independently?

This seems to be creating more confusion than it is worth.  To be clear, I have a long series of pull requests coming which generally must be merged in order, one at a time.  Each request should be safe to merge without breaking the exult build, so we don't need to merge them all in one batch.

My intent was to post a small collection of closely-related patches as pull requests so that you can look them over in context, and potentially merge more than one per day if they look good.

I have each patch staged on a separate/dedicated branch, but each branch is rooted off the previous one.  When I create a PR for each branch, the first one has exactly one patch in it, but the later ones show the cumulative patch series.  The intent is to just review the final patch in each PR and disregard the cumulative earlier patches.

I did this for the first set of PRs and I believe @DominusExult was able to merge several of them successfully without me having to rebase; git sorted out the common history during the merge and they showed up correctly in master.

If you would prefer, I can stick to just submitting one PR at a time to avoid further confusion.

ceckak:
Ok - I got a start on refactoring to add reference variables to reduce the number of code changes.  May take me a few days to wade through all the changes; will post an update when I'm done.

ceckak:
Done.  I've refactored the changes to bias heavily toward minimizing the changes in the commit by using temporary references.

marzojr (approved):

PR #182, opened 2022-04-18
Use U7open_in() to check if files exist

ceckak:
A susbsequent patch will introduce new file access abstractions which do not support calling `stat()`.  Switch to speculatively opening files for reading to test if they exist as a preliminary cleanup step.

### DIFF
--- a/audio/midi_drivers/MT32EmuMidiDriver.cpp
+++ b/audio/midi_drivers/MT32EmuMidiDriver.cpp
@@ -94,15 +94,15 @@ int MT32EmuMidiDriver::open() {
 		FileStream part1;
 		FileStream part2;
 		if (openROMFile(part1, "MT32A.BIN", false) && openROMFile(part2, "MT32B.BIN", false)) {
-			std::ofstream out;
-			if (U7open(out, "<SAVEHOME>/data/MT32_CONTROL.ROM", false)) {
+			auto pOut = U7open_out("<SAVEHOME>/data/MT32_CONTROL.ROM", false);
+			if (pOut) {
+				auto& out = *pOut;
 				const Bit8u *data1 = part1.getData();
 				const Bit8u *data2 = part2.getData();
 				for (size_t ii = 0; ii < std::min(part1.getSize(), part2.getSize()); ii++) {
 					out.put(static_cast<char>(data1[ii]));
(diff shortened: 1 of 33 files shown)


Dragon-Baroque:
Please, in `files/utils.cc` why do you need to test whether `istream_factory` is non NULL in `U7Remove` ?

ceckak:
> Please, in `files/utils.cc` why do you need to test whether `istream_factory` is non NULL in `U7Remove` ?

You're talking about line 387?
```
if (in && in->good() && !in->fail()) {
```

I don't recall if I had a concrete case where I needed it, but in theory, `istream_factory()` could fail and return `nullptr`, so I think I was just being defensive here to avoid a potential crash.  This is a `void` function so I can't propagate an error, but I could log an error or throw an exception if you prefer.

Dragon-Baroque:
No, I was refering to
```
	if (istream_factory) {
		in = istream_factory(name.c_str(), std::ios_base::in);
	} else {
		in = std::make_unique<std::ifstream>(name.c_str(), std::ios_base::in);
	}
```

ceckak:
> No, I was refering to
> 
> ```
> 	if (istream_factory) {
> 		in = istream_factory(name.c_str(), std::ios_base::in);
> 	} else {
> 		in = std::make_unique<std::ifstream>(name.c_str(), std::ios_base::in);
> 	}
> ```

Ah - I see.  This was introduced to address some early feedback I got from @DominusExult.  I originally was depending on the SDL-backed streams by default, and this forced all the standalone utilities that interact with files to take an SDL dependency.  To address that, I introduced the factory so that the SDL dependency can be injected by `exult.cc`, and if no factory has been provided, I fall back to standard file access.  This enables all the standalone utilities to continue to work without having to provide a factory or take an SDL dependency.

marzojr (changes requested):

marzojr:
Other than those, looks good.

marzojr:
Hm. These notes should be for #181, for some reason the notifications in GH were on the wrong order.

ceckak:
> Other than those, looks good.

Thanks for reviewing.  Suggested whitespace changes are implemented in #181

marzojr (approved):

marzojr (approved):

PR #183, opened 2022-04-18
Use SDL_RWops for file access where possible

ceckak:
Some platforms have nontraditional filesystem access and SDL has already provided an abstraction layer for them in the form of [SDL_RWops](https://wiki.libsdl.org/SDL_RWops).  Rather than introduce platform-specific code into exult to account for platform filesystem differences, leverage SDL's abstraction as the primary interface for interacting with the filesystem.

SDL_rwops was officially introduced in SDL 2.0.10, so this patch also updated the minimum SDL version to that level, and fixes a minor bug in the SDL version checking in `acinclude.m4`.

Note that there is one glaring deficiency in `SDL_RWops`: it does not provide an api for interacting with directories.  So while this change works for reading and writing individual files, it does not help with searching directories to see what they contain, so we end up falling back to legacy code that bypasses SDL, which might not work on all platforms.  This might cast doubt on whether there is value in switching to `SDL_RWops`, but the only place exult currently queries directories is when detecting mods.  This means mods need to live in a traditional filesystem location, but all other content can be placed in platform-specific locations.  This works out well for android where some of the data files can be bundled directly into the apk, while mods are installed into the device's filesystem where they can be detected.

### DIFF
--- a/files/sdlrwopsstreambuf.cc
+++ b/files/sdlrwopsstreambuf.cc
@@ -1,0 +1,281 @@
+/*
+ *  Copyright (C) 2021-2022  The Exult Team
+ *
+ *  This program is free software; you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
+ *  the Free Software Foundation; either version 2 of the License, or
+ *  (at your option) any later version.
+ *
+ *  This program is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License
+ *  along with this program; if not, write to the Free Software
+ *  Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA 02111-1307, USA.
+ */
+
+#include "sdlrwopsstreambuf.h"
+
+#include <SDL.h>
+
+#include <cstring>
+#include <memory>
+
+// Partly derived from libc++'s basic_filebuf
+// https://github.com/llvm-mirror/libcxx/blob/master/include/fstream
+//   Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+//   See https://llvm.org/LICENSE.txt for license information.
+//   SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+static constexpr auto NO_OPENMODE = static_cast<std::ios_base::openmode>(0);
+
+SdlRwopsStreambuf::SdlRwopsStreambuf()
+		: m_context(nullptr), m_openMode(NO_OPENMODE),
+		  m_currentMode(NO_OPENMODE) {
+	setg(nullptr, nullptr, nullptr);
+	setp(nullptr, nullptr);
+}
+
+SdlRwopsStreambuf::~SdlRwopsStreambuf() {
+	try {
+		close();
+	} catch (...) {
+	}
+}
+
+bool SdlRwopsStreambuf::is_open() const {
+	return m_context;
+}
+
+SdlRwopsStreambuf*
+		SdlRwopsStreambuf::open(const char* s, std::ios_base::openmode mode) {
+	if (m_context) {
+		return nullptr;
+	}
+
+	const char* mdstr   = nullptr;
+	int         modeval = mode & ~std::ios_base::ate;
+	switch (modeval) {
+	case std::ios_base::out:
+	case std::ios_base::out | std::ios_base::trunc:
+		mdstr = "w";
+		break;
+	case std::ios_base::out | std::ios_base::app:
+	case std::ios_base::app:
+		mdstr = "a";
+		break;
+	case std::ios_base::in:
+		mdstr = "r";
+		break;
+	case std::ios_base::in | std::ios_base::out:
+		mdstr = "r+";
+		break;
+	case std::ios_base::in | std::ios_base::out | std::ios_base::trunc:
+		mdstr = "w+";
+		break;
+	case std::ios_base::in | std::ios_base::out | std::ios_base::app:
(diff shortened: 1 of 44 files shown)


ceckak:
Fixed the link error on windows (it wasn't pulling in the  new `sdlrwops*` sources).  Presumably the same issue exists for ios, but I'm not set up to edit the ios project to add them.  Might need some help there if we want to include that with this commit, otherwise we can circle back and fix the ios build separately after the dust settles from the android merge.

DominusExult:
I can quickly fix it for the iOS port once this is merged.

Can anyone please review this series of pull requests? I'm really not equipped to do this (other than reporting that a qick test seemed fine). @marzojr @drcode1 @Dragon-Baroque

Dragon-Baroque:
Sure willing to help you, @DominusExult.
Now what is the most efficient way to bring a pull request into my local Git trees from a fork [ `ceckak/exult.git` ] that my Git installed trees don't know about [ I have locally installed reference `Exult/exult.git` and my own fork `Dragon-Baroque/exult.git` ] ?

DominusExult:
Thanks! Last time I just checked out the fork into a new folder, tried there and then merged from the github page.
But I think you could also add this fork as a remote. But as ceckak changes his fork "destructively" at times the first way might be better to keep your working copy clean and unbroken ;)

ceckak:
> as ceckak changes his fork "destructively" at times

Guilty as charged.  I've been managing this for as a way to distribute a patch series that I'm continuing to refine until they get merged, but I can pivot to preserving history if that makes things easier for you guys.

DominusExult:
I'm fine with it ;)
I just noticed it when I suddenly had an abandoned fork with my iOS fixes for exactly this issue a while ago ;)

Dragon-Baroque:
> but I can pivot to preserving history if that makes things easier for you guys.

Thanks for the offer, but not needed. I did exactly the same in my own Pull Requests. Only I am now on the receiving side, and @DominusExult essentially stated that I should take caution if I used the `remote` strategy as compared to the `clone` strategy.

ceckak:
> also his PRs always contains stuff from other PRs. for example this one contains the sdl_main commit that was filed as a separate PR. if this one here is merged as-is, there will be 2 commits "Remove sdl_main.cc wrapper"
which is not only confusing to the reader of the git history, but it will also break automated tools assuming a flat/continuous history.

I would love to fix that if someone can tell me how.  I have a series of 46 patches which largely depend on each other and need to be merged in series.  I didn't want to submit a single PR with all of them since it makes it hard for you to review them independently, but it also seemed helpful to submit several related PRs at a time.  I have yet to find a way to do that which doesn't show the duplication you're referring to.  I am happy to rebase the later PRs as the earlier ones get merged though, which eliminates the duplication while retaining the comment thread on the PR.

rofl0r:
> I would love to fix that if someone can tell me how.

there's only two ways to achieve that i'm aware of:

- a single PR containing all patches
- a single PR at a time, wait for merge, file next

ceckak:
Ok - that's consistent with what I've been seeing.  Feel free to ignore the later PRs if they aren't useful for context; I will rebase them as we merge from the head of the queue.

Dragon-Baroque:
Have you noticed that the compilation checks on Linux failed ?
This is not your fault, but you could fix it :
- You require SDL 2.0.10 in `acinclude.m4` [ was 2.0.6 before the commit ],
- The Ubuntu Linux installs 2.0.10 _exactly,_
- The version test line 72 in `acinclude.m4` is wrong, it uses a `-gt` for the Patchlevel, should use a `-ge.`

Could you fix this and retry ? That should pass the Linux compilation tests.

ceckak:
> Have you noticed that the compilation checks on Linux failed ? This is not your fault, but you could fix it :
> 
> * You require SDL 2.0.10 in `acinclude.m4` [ was 2.0.6 before the commit ],
> * The Ubuntu Linux installs 2.0.10 _exactly,_
> * The version test line 72 in `acinclude.m4` is wrong, it uses a `-gt` for the Patchlevel, should use a `-ge.`
> 
> Could you fix this and retry ? That should pass the Linux compilation tests.

Thanks for chasing that down.  I had noticed the build failures, but hadn't realized the SDL version change was triggering it.  I'll add a fix for the version test as you've suggested.

files/sdlrwopsstreambuf.cc
```diff
@@ -0,0 +1,281 @@
...
+			return traits_type::eof();
+		setp(pb_save, epb_save);
+	}
+	return traits_type::not_eof(c);
+}
+
+typename SdlRwopsStreambuf::pos_type SdlRwopsStreambuf::seekoff(
+		off_type off, std::ios_base::seekdir dir, std::ios_base::openmode) {
+	int width = 1;
+	if (!m_context || (width <= 0 && off != 0) || sync())
```
marzojr, line 223: We already know width is 1 here, so the check `width <= 0` is always false (see [LGTM 1](https://lgtm.com/projects/g/exult/exult/snapshot/a28cf7660fd554c904f32373c65983cdbc5fb585/files/files/sdlrwopsstreambuf.cc#xf56c03303f434649:1).

files/sdlrwopsstreambuf.cc
```diff
@@ -0,0 +1,281 @@
...
+	case std::ios_base::cur:
+		whence = RW_SEEK_CUR;
+		break;
+	case std::ios_base::end:
+		whence = RW_SEEK_END;
+		break;
+	default:
+		return pos_type(off_type(-1));
+	}
+	if (SDL_RWseek(m_context, width > 0 ? width * off : 0, whence) == -1)
```
marzojr, line 239: We already know width is 1 here, so the ternary operator can be replaced by `width * off` (see [LGTM 2](https://lgtm.com/projects/g/exult/exult/snapshot/a28cf7660fd554c904f32373c65983cdbc5fb585/files/files/sdlrwopsstreambuf.cc#x6edf1c4e1249dcaa:1).

marzojr (changes requested):

marzojr (approved):

PR #184, opened 2022-04-18
Use U7 file I/O for unzip

ceckak:
The unzip code is hard-coded to to directly access the filesystem, which ends up bypassing the logic in in U7 file I/O for game folders and handling nonstandard directory structures and filesystems.  This patch modifies the unzip code to use the U7 file I/O abstraction.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -52,6 +52,12 @@ EXULTSOURCES =	\
 	exultmenu.cc	\
 	exultmenu.h	\
 	exult_constants.h	\
+	files/sdlrwopsstreambuf.cc \
+	files/sdlrwopsstreambuf.h \
+	files/sdlrwopsistream.cc \
+	files/sdlrwopsistream.h \
+	files/sdlrwopsostream.cc \
+	files/sdlrwopsostream.h \
 	fnames.h	\
 	frameseq.h	\
 	game.cc		\
--- a/Makefile.common
+++ b/Makefile.common
@@ -13,6 +13,9 @@ MAIN_OBJS:=actions.o \
 	effects.o \
 	exult.o \
 	exultmenu.o \
+	files/sdlrwopsstreambuf.o \
+	files/sdlrwopsistream.o \
+	files/sdlrwopsostream.o \
 	game.o \
 	gameclk.o \
 	gamedat.o \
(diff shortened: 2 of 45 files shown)


Dragon-Baroque:
By replacing plain `fopen` by `U7fopen` and thus by `U7open_in`, you enforce the case insensitivity on the file names and the file paths, even on platforms that use case sensitive file systems. 
This is certainly acceptable from the Ultima 7 DOS files. But is this acceptable for the ZIP files ?

This said, I have two nitpicks :
- There are a lot of `CASESENSITIVITY` macros in `unzip.cc ` which control the comparison functions in the ZIP directory. I think you should enforce them to `insensitive` to be consistent with the change to `U7fopen`.
- You should add a few comments on the top and the bottom of your changes in `unzip.cc` to comply with the requirements of the original author as stated in the beginning of that file, identifying these changes as applying to Exult. Or we may take the stronger position that the ZIP related code _does not belong to Exult_, and that is why it is in its own subdirectory, and therefore we do not alter it. What about this, @DominusExult ?

DominusExult:
I think we may be good with the 

> Modified by the Exult Team. Nov 10th 2001 - 2022
And maybe an additional 

> Modified by the Ken Ceckak. !Date!

if we need to annotate everything we may need to take a much closer look and add comments as we have some uncommented changes there ;)

ceckak:
@Dragon-Baroque - good point about the change in case sensitivity.  I believe the change to `U7fopen` here *only* affects the case sensitity of the zipfile name/path itself, not the way unzip interacts with the contents of the zip file.  If I'm reading the code correctly, the CASESENSITIVITY macros relate to the handling of the contents of the zip file.  Given that distinction, you still think we should switch to case-insensitive for the zipfile contents?

In general, my thinking with this change is that U7 has a standardized way of finding files in the filesystem, and that it should use that way everywhere, including when locating zip files to open.  This was necessary on Android as some of the paths are in unusual locations.  I'd have to go back and do some testing to verify which ones in particular.  If you don't like using `U7fopen` in unzip, I can probably find some other way to handle it outside of the unzip code.

I'll add some additional comments to the code to highlight the modifications as suggested.

marzojr (approved):

marzojr (approved):

PR #187, opened 2022-04-21
Support C++ compiler override in configure.ac

Dragon-Baroque:
Commit 1 of 1 :
    ./configure.ac
      Complete the handling of option --with-cxx
      Handle the OVERRIDE_CXX environment variable
      Disable the GIMP plugin build if pkg-config is not installed
      Remove a few spurious spaces and tabs
      List a larger set of option settings, including the compiler

For the replacement commit : remove the display of ( nonexistent ) `enable_macosx_studio_support`

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -60,7 +60,6 @@ case "$host_os" in
 		CXXFLAGS="$CXXFLAGS -D_USE_MATH_DEFINES"
 		SYSLIBS="-luuid -lole32 -lwinmm -lstdc++ -lws2_32"
 		ICON_FILE="win32/exultico.o"
-		enable_timidity="no"
 		;;
 	cygwin* )
 		WINDOWING_SYSTEM="-D_WIN32"
@@ -69,7 +68,6 @@ case "$host_os" in
 		CXXFLAGS="$CXXFLAGS -mno-cygwin"
 		SYSLIBS="-lwinmm"
 		ICON_FILE="win32/exultico.o"
-		enable_timidity="no"
 		;;
 	openbsd* )
 		WINDOWING_SYSTEM="-DXWIN"
@@ -132,12 +130,6 @@ AM_CONDITIONAL(ANDROID, test x$ARCH = xandroid)
 # ---------------------------------------------------------------------
 
 AC_PROG_AWK
-
-AC_ARG_WITH(cxx, AS_HELP_STRING([--with-cxx=COMMAND], [Explicitly specify the C++ compiler to use]),[WITHCXX="$withval"],[WITHCXX="$CXX"])
-
-if test "x$WITHCXX" != "x"; then
-	cxx_compilers="$WITHCXX"
-fi
 
 AC_PROG_CXX([$cxx_compilers])
 
@@ -869,7 +861,6 @@ AC_MSG_CHECKING([whether to build only the main program])
 if test x$enable_tools = xno; then
 	AC_MSG_RESULT(yes)
 	AM_CONDITIONAL(BUILD_TOOLS, false)
-	enable_gtk_interface=no
 	enable_gimp_plugin=no
 	enable_compiler=no
 else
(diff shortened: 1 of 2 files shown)


configure.ac
```diff
@@ -133,10 +133,21 @@ AM_CONDITIONAL(ANDROID, test x$ARCH = xandroid)
 
 AC_PROG_AWK
 
-AC_ARG_WITH(cxx, AS_HELP_STRING([--with-cxx=COMMAND], [Explicitly specify the C++ compiler to use]),[WITHCXX="$withval"],[WITHCXX="$CXX"])
+AC_MSG_CHECKING([for compiler override by OVERRIDE_CXX environment variable])
+if test "x$OVERRIDE_CXX" != "x"; then
+	AC_MSG_RESULT([yes, with $OVERRIDE_CXX])
```
marzojr, line 138: It might be better to use CXX directly and change the ci-linux workflow to use it instead, as @rofl0r mentioned.

marzojr (changes requested):

configure.ac
```diff
@@ -1183,6 +1200,31 @@ echo Build HQ4X scaler.......... : $enable_hq4x
...
+echo enable_zip_support......... : $enable_zip_support
+echo
+echo enable_pedantic_errors..... : $enable_pedantic_errors
+echo enable_sanitizers.......... : $enable_sanitizers
+echo enable_sdl_parachute....... : $enable_sdl_parachute
+echo enable_static_libs......... : $enable_static_libs
+echo enable_warning_errors...... : $enable_warning_errors
+echo
+echo enable_timidity............ : $enable_timidity
+echo enable_gtk_interface....... : $enable_gtk_interface
```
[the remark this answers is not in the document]
Dragon-Baroque: Most variables are only internal replicas of the `--with` and other options. I shall transfer their description.

The two at the end on the other hand look like spelling mistakes or not completed code...

configure.ac
Dragon-Baroque: I hane taken a look into the generated code of the compiler choice macros 
```
AC_PROG_CXX([$cxx_compilers])·
·
AC_LANG([C++])·
```
It looks to me as if `$CXX` is indeed taken when set, but then it acts as a very quick exit bypassing all internal checks.

I would agree with the direct use of `$CXX` if we were using simple `make`, but we are not, we are using the Autotools and set the compiler and its options in a rather complex way. To say it otherwise, I look at the `$CXX` or the `$CXXFLAGS` as the output of the Autotools compiler macros and the input to the Makefiles, not the input of the Autotools.

Then, I felt that if the `ci-linux` workflow was using `$OVERRIDE_CXX` and not `$CXX` directly, which would have been taken immediately and would have provided CLang support to Linux Exult from the beginning, it must have been a good reason ... Am I wrong ?

marzojr: Thinking back, I think that I forgot to add either `--with-cxx=$(which $OVERRIDE_CXX)` or `CXX=$(which $OVERRIDE_CXX)` as arguments to `./configure`, the later being the "recommended" way of specifying a compiler in autotools. On a personal project which uses a similar setup (but with cmake instead of autotools), I export CXX before calling cmake to "configure", which is the recommended way of doing that with cmake.

For what is worth, I think that the logic of AC_PROG_CXX skipping most internal checks when you specify a compiler by hand is that when you do this, you are basically assuming the responsibility of ensuring that the compiler can do the job.

Dragon-Baroque: Then why not simplify into :

1. `configure.ac` : remove `--with-cxx` and the handling of `$OVERRIDE_CXX` and let `make` / the Autotools acquire `CXX` either from the environment variable `$CXX` or from the `CXX=` option to `./configure`,
2. `ci-linux` : replace OVERRIDE_CXX by CXX and either stop there or pass `CXX=${CXX}` to './configure` ?

marzojr: That works, yes.

configure.ac
```diff
@@ -1182,6 +1188,36 @@ echo Build HQ3X scaler.......... : $enable_hq3x
...
+echo
+echo Compilation :
+echo Enable compiler sanitizers. : $enable_sanitizers
+echo Enable SDL parachute....... : $enable_sdl_parachute
+echo Build with static libraries : $enable_static_libs
+echo Build libexult library..... : $enable_libexult
+echo Turn all warnings to errors : $enable_warning_errors
+echo Turn pedantics to errors... : $enable_pedantic_errors
+echo
+echo enable_timidity............ : $enable_timidity
```
marzojr, line 1218: I have been looking over; and more than 17 years ago, this variable was used as a default for Timidity support if not specified. There was a merge from Pentagram audio code (in d8771b6 ) which renamed the variable in most places from `enable_timidity` to `enable_timidity_midi`, and the defaults were no longer used. It can probably go away, since it has not caused any issues in the past 17 years.

configure.ac
```diff
@@ -1182,6 +1188,36 @@ echo Build HQ3X scaler.......... : $enable_hq3x
...
+echo enable_gtk_interface....... : $enable_gtk_interface
```
marzojr, line 1219: This variable, on the other hand, was only ever set to 'no' in one spot and not used for anything, ever, as far back as we have a configure.ac. It probably can go away too.

marzojr (approved):

marzojr (approved):

PR #191, opened 2022-04-24
Fix ogg/vorbis stuttering on Android

ceckak:
The existing 20Hz buffering for ogg/vorbis playback isn't sufficient on android.  It results in stuttering when using digital music on both an emulated android device on a laptop and on two native android devices.  Increasing to 5Hz buffering.

Note: This value was chosen based on empirical testing on a Kindle Fire tablet.  6Hz and above resulted in stuttering with digital music enabled.

### DIFF
--- a/audio/AudioMixer.cc
+++ b/audio/AudioMixer.cc
@@ -85,9 +85,10 @@ id_counter(0), device(nullptr)
 	desired.callback = sdlAudioCallback;
 	desired.userdata = this;
 
-	// Set update rate to 20 Hz, or there abouts. This should be more then adequate for everyone
+	// Set update rate to 5 Hz, or there abouts. This should be more than adequate for everyone
+	// Note: setting this to 1 Hz (/1) causes Exult to hang on MacOS.
 	desired.samples=1;
-	while(desired.samples<=desired.freq/30) desired.samples<<=1;
+	while(desired.samples<=desired.freq/5) desired.samples<<=1;
 
 	// Open SDL Audio (even though we may not need it)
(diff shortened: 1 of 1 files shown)


audio/AudioMixer.cc
```diff
@@ -85,9 +85,10 @@ id_counter(0), device(nullptr)
 	desired.callback = sdlAudioCallback;
 	desired.userdata = this;
 
-	// Set update rate to 20 Hz, or there abouts. This should be more then adequate for everyone
+	// Set update rate to 2 Hz, or there abouts. This should be more than adequate for everyone
+	// Note: setting this to 1 Hz (/1) causes Exult to hang on MacOS.
 	desired.samples=1;
-	while(desired.samples<=desired.freq/30) desired.samples<<=1;
+	while(desired.samples<=desired.freq/2) desired.samples<<=1;
```
marzojr, line 91: Does it stutter with a higher value than 2? Like 10 or something? I am concerned about increased latency at the start of each audio track.

ceckak: I can do some experimenting to find a boundary.  I wasn't sure what the implications where of taking this to an extreme, but if it increases latency, I can try to find somewhere in the middle.  Will try to do some experimenting tonight.

ceckak: Did some binary searching with frequencies on my android tablet and settled on 5Hz as the largest stable value.  Anything above that stutters.  If you're concerned that this is still to aggressive for a universal default, I can ifdef it for 
Android.

marzojr (approved):

PR #196, opened 2022-04-24
Fixes few GCC warnings

dmorilha:
Hello, this PR fixes few of the warnings I see while compiling exult with gcc-11.2 on arch linux. Hopefully nothing too controvertial.

### DIFF
--- a/audio/midi_drivers/XMidiFile.cpp
+++ b/audio/midi_drivers/XMidiFile.cpp
@@ -723,16 +723,17 @@ void XMidiFile::ApplyFirstState(first_state &fs, int chan_mask)
 		patch->time = 0;
 		bank->time = 0;
 
-		if (reverb && do_chorus) reverb->next = chorus;
-		else if (reverb) reverb->next = bank;
-		if (chorus) chorus->next = bank;
+		if (reverb)
+			if (chorus) reverb->next = chorus;
+			else reverb->next = bank;
+		else if (chorus) chorus->next = bank;
 		bank->next = vol;
 		vol->next = pan;
 		pan->next = patch;
 
 		patch->next = list;
-		if (do_reverb) list = reverb;
-		else if (do_chorus) list = chorus;
+		if (reverb) list = reverb;
+		else if (chorus) list = chorus;
 		else list = bank;
 	}
 }
--- a/audio/midi_drivers/XMidiFile.cpp
+++ b/audio/midi_drivers/XMidiFile.cpp
@@ -723,9 +723,9 @@ void XMidiFile::ApplyFirstState(first_state &fs, int chan_mask)
 		patch->time = 0;
 		bank->time = 0;
 
-		if (do_reverb && do_chorus) reverb->next = chorus;
-		else if (do_reverb) reverb->next = bank;
-		if (do_chorus) chorus->next = bank;
+		if (reverb && do_chorus) reverb->next = chorus;
+		else if (reverb) reverb->next = bank;
+		if (chorus) chorus->next = bank;
 		bank->next = vol;
 		vol->next = pan;
 		pan->next = patch;
--- a/gamedat.cc
+++ b/gamedat.cc
@@ -482,7 +482,7 @@ void Game_window::write_saveinfo() {
 
 			char name[18];
 			std::string namestr = npc->get_npc_name();
-			strncpy(name, namestr.c_str(), 18);
+			memcpy(name, namestr.c_str(), 18);
 			out.write(name, 18);
 			out.write2(npc->get_shapenum());
 
@@ -1065,7 +1065,7 @@ bool Game_window::save_gamedat_zip(
 		ofstream out;
 		char title[0x50];
 		memset(title, 0, 0x50);
-		std::strncpy(title, savename, 0x50);
+		memcpy(title, savename, 0x50);
 		U7open(out, fname);
 		out.write(title, 0x50);
 		out.close();
--- a/gamedat.cc
+++ b/gamedat.cc
@@ -480,9 +480,9 @@ void Game_window::write_saveinfo() {
 			else
 				npc = get_npc(party_man->get_member(i - 1));
 
-			char name[18];
+			char name[18] = {0};
 			std::string namestr = npc->get_npc_name();
-			memcpy(name, namestr.c_str(), 18);
+			strcpy(name, namestr.substr(0, 17).c_str());
 			out.write(name, 18);
 			out.write2(npc->get_shapenum());
 
@@ -1063,9 +1063,8 @@ bool Game_window::save_gamedat_zip(
 	// Name
 	{
 		ofstream out;
-		char title[0x50];
-		memset(title, 0, 0x50);
-		memcpy(title, savename, 0x50);
+		char title[0x50] = {0};
+		strncpy(title, savename, 0x4F);
 		U7open(out, fname);
 		out.write(title, 0x50);
 		out.close();
--- a/gamemgr/bggame.cc
+++ b/gamemgr/bggame.cc
@@ -1934,16 +1934,16 @@ void BG_Game::end_game(bool success) {
 						// if(year > 0) sprintf(buffer,"%d year(s) , ",year);
 						// message = buffer;
 						if (month == 1)
-							sprintf(buffer + strlen(buffer), "%d month", month);
+							strcat(buffer, "1 month");
 						else if (month > 1)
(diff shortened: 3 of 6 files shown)


gamedat.cc
```diff
@@ -482,7 +482,7 @@ void Game_window::write_saveinfo() {
 
 			char name[18];
 			std::string namestr = npc->get_npc_name();
-			strncpy(name, namestr.c_str(), 18);
+			memcpy(name, namestr.c_str(), 18);
```
marzojr, line 485: Ack, no. Undefined behavior here whenever `namestr.size() < 17`.  A better solution is to do away with `name` variable, replace the `strncpy` call by `namestr.resize(18, '\0')` , and use `namestr.c_str()` instead of `name` in the call to write.

gamedat.cc
```diff
@@ -1065,7 +1065,7 @@ bool Game_window::save_gamedat_zip(
 		ofstream out;
 		char title[0x50];
 		memset(title, 0, 0x50);
-		std::strncpy(title, savename, 0x50);
+		memcpy(title, savename, 0x50);
```
marzojr, line 1068: Once again, undefined behavior here whenever `strlen(savename) < 0x4f`. In this case, the only solution is doing something like `memcpy(title, savename, std::max(0x50, strlen(savename)))`.

gamemgr/bggame.cc
```diff
@@ -1935,41 +1934,38 @@ void BG_Game::end_game(bool success) {
 						// if(year > 0) sprintf(buffer,"%d year(s) , ",year);
 						// message = buffer;
 						if (month == 1)
-							sprintf(buffer, "%s%d month", message, month);
+							sprintf(buffer + strlen(buffer), "%d month", month);
```
marzojr, line 1937: Maybe strncat with `"1 month"`?

gamemgr/bggame.cc
```diff
@@ -1935,41 +1934,38 @@ void BG_Game::end_game(bool success) {
...
 						else if (month > 1)
-							sprintf(buffer, "%s%d months", message, month);
-						message = buffer;
+							sprintf(buffer + strlen(buffer), "%d months", month);
 
 						// add ampersand only if month(s) and there is more to display.
 						if (month > 0 && (day != 0 || hour != 0))
-							sprintf(buffer, "%s & ", message);
-						message = buffer;
+							sprintf(buffer + strlen(buffer), " & ");
```
marzojr, line 1943: Maybe use strncat?

gamemgr/bggame.cc
```diff
@@ -1935,41 +1934,38 @@ void BG_Game::end_game(bool success) {
...
 
 						if (day == 1)
-							sprintf(buffer, "%s%d day", message, day);
+							sprintf(buffer + strlen(buffer), "%d day", day);
```
marzojr, line 1946: Maybe strcat with `"1 day"`?

gamemgr/bggame.cc
```diff
@@ -1935,41 +1934,38 @@ void BG_Game::end_game(bool success) {
...
 						else if (day > 1)
-							sprintf(buffer, "%s%d days", message, day);
+							sprintf(buffer + strlen(buffer), "%d days", day);
 						// if no days, display hours(this would only happen on exactly 1,2,3 etc months)
 						// Here so the player doesnt think we didn't track the hours/days.
 						// so 112 days at 2am would display "4 months & 2 hours", 113 days at 2am would display "4 months & 1 day"
 						else if (day == 0 && hour == 1)
-							sprintf(buffer, "%s%d hour", message, hour);
+							sprintf(buffer + strlen(buffer), "%d hour", hour);
```
marzojr, line 1953: Maybe strcat with `"1 hour"`?

gamemgr/bggame.cc
```diff
@@ -1935,41 +1934,38 @@ void BG_Game::end_game(bool success) {
...
 						else if (day == 0 && hour > 1)
-							sprintf(buffer, "%s%d hours", message, hour);
+							sprintf(buffer + strlen(buffer), "%d hours", hour);
 
 						// in the remote chance a player finishes on exactly 0 hours, 0 days and X month(s)
 						if (day == 0 && hour == 0)
-							sprintf(buffer, "%s & 0 days", message);
-						message = buffer;
+							sprintf(buffer + strlen(buffer), " & 0 days");
```
marzojr, line 1959: Maybe strcat with `" & 0 days"`?

gumps/Newfile_gump.cc
```diff
@@ -919,7 +918,7 @@ void Newfile_gump::LoadSaveGameDetails() {
 			npc = gwin->get_npc(partyman->get_member(i - 1));
 
 		std::string namestr = npc->get_npc_name();
-		strncpy(cur_party[i].name, namestr.c_str(), 18);
+		memcpy(cur_party[i].name, namestr.c_str(), 18);
```
marzojr, line 921: More undefined behavior here whenever `namestr.size() < 17`. You need to call `namestr.resize(18, '\0')` before this line.

audio/midi_drivers/XMidiFile.cpp
```diff
@@ -723,9 +723,9 @@ void XMidiFile::ApplyFirstState(first_state &fs, int chan_mask)
 		patch->time = 0;
 		bank->time = 0;
 
-		if (do_reverb && do_chorus) reverb->next = chorus;
-		else if (do_reverb) reverb->next = bank;
-		if (do_chorus) chorus->next = bank;
+		if (reverb && do_chorus) reverb->next = chorus;
+		else if (reverb) reverb->next = bank;
+		if (chorus) chorus->next = bank;
```
marzojr, line 728: In my opinion, this change is pointless, as the pointers can never be null if the booleans are true. But in any case, you might want to also replace do_chorus in line 726 by chorus for consistency.

gumps/Gamemenu_gump.cc
```diff
@@ -54,7 +54,6 @@ static const char *displaytext = "Game Display";
 static const char *enginetext = "Game Engine";
 static const char *inputtext = "Game Input";
 #ifndef __IPHONEOS__
-static const char *quitmenutext = "Quit to Menu";
```
marzojr, line 57: Please don't remove this line. Comment it if you must, but leave it in.

marzojr (changes requested):

rofl0r:
counter proposal: rather than introducing a ton of gratuitous and potentially buggy changes to make OP happy by seeing a silent build, just add -Wno-whatever.

gamedat.cc
```diff
@@ -480,9 +480,9 @@ void Game_window::write_saveinfo() {
 			else
 				npc = get_npc(party_man->get_member(i - 1));
 
-			char name[18];
+			char name[18] = {0};
 			std::string namestr = npc->get_npc_name();
-			memcpy(name, namestr.c_str(), 18);
+			strcpy(name, namestr.substr(0, 17).c_str());
```
marzojr, line 485: This breaks the case of a name with length 18, which is valid and works in the previous version.

gumps/Newfile_gump.cc
```diff
@@ -918,7 +918,7 @@ void Newfile_gump::LoadSaveGameDetails() {
 			npc = gwin->get_npc(partyman->get_member(i - 1));
 
 		std::string namestr = npc->get_npc_name();
-		memcpy(cur_party[i].name, namestr.c_str(), 18);
+		strcpy(cur_party[i].name, namestr.substr(0, 17).c_str());
```
marzojr, line 921: This breaks the case of a name with length 18, which is valid and works in the previous version.

gamedat.cc
dmorilha: so name has to be 19 bytes long and everything would work just fine.

dmorilha:
Exult is a great project for those wanting to learn more about games. It not only revamps the precursor of open world games, but it does adding to the original game! Nonetheless it could use some code modernization or at least refactoring based on a standard for a better reading experience.

Over the weekend I was playing with the idea of adding / or continuing to add joystick support, not sure how I feel about contributing it back...

DominusExult:
Don't be disheartened by the 1st hurdle ;)
More joystick control would be appreciated (we have movement working nicely).
And yes, Exult could use code refactoring, for example the SDL event polling for keyboard/mouse/joystick is repeated several times.

But at the same time, Exult is fairly sized project and you just experienced how easy it is to break things. Compiler warnings are not always completely true and the obvious way to silence them might overlook the reason it is done a certain way and may break the game or other systems.

PR #205, opened 2022-05-01
Add a simple launcher activity

ceckak:
This patch series adds the basic android launcher which can start up the exult engine.


android/app/src/main/cpp/exult_android_main.cc
```diff
@@ -30,6 +31,8 @@ ExultAndroid_main(int argc, char* argv[]) {
 	auto                 ndk_cout = std::cout.rdbuf(&exult_cout);
 	auto                 ndk_cerr = std::cerr.rdbuf(&exult_cerr);
 
+        auto result = SDL_main(argc, argv);
```
marzojr, line 34: Mixing tabs and spaces.

marzojr (changes requested):

android/app/src/main/cpp/exult_android_main.cc
ceckak: done

marzojr (approved):

PR #224, opened 2022-06-23
Vita Support for Exult

einmueller:
This is a not-so-big patch to compile and play Exult on Sony Vita.

The left joystick is for moving, the right one works as mouse with left-shoulder button as left mouse button.
Keyboard Shortcuts are configurable. A README.Vita is included.
Digital Music works, but the vita is too slow for this. With Midi-Sound the game runs pretty fine.

Some glibc-functions had to be reimplemented and the SDL-Joystick-Support to be fully activated. 
(The 2nd part could perhaps also be interesting for casual gamers with a gamepad on their pc, but
is only activated on vita so far)

The whole patch gets activated when compiling with "-DVITA", it needs the vita-sdk to generate 
a vita package. Precompiled packages are on my github-site if you wish to link them back to
the exult web-page. All sources are in the vita-subfolder. 

Greetings an thanks for all your hard work over the years!!!


rofl0r:
congratulations, you get the prize for "how not to make a pull requests"

- use lots of commits with meaningless commit messages
- even use duplicate commit messages with typos in them
- mess up commit history with merge commits
- add tons of ifdefs
- add lots of commented out code
- fuck up whitespace all over the place
- replace text in original README with reference to a fork on another repo

:clap: :clap::clap:

DominusExult:
@rofl0r please don't be THAT guy. We don't want to antagonize contributors in that way.

@einmueller he is right that your pull request needs a lot of tidy up, by squashing commits into smaller logical steps, cleaner code (no commented out code, whitespaces...).
But thanks for taking your time to work on adding this port and trying to merge with us. Please disregard the tone of rofl0r.

wench:
My thoughts: The changes to readme are unacceptable to me, Removes things without reason. Yes that stuff may not be wanted in a fork but merging those changes into our repository makes no sense. 

This PR turns exult into the fork instead of only applying the specific changes needed for platform support.

Code should be written as portable as possible without ifdefs relying on SDL as much as possible to handle platform stuff. Many of the change themselves seem like they should be done in a completely platform portable way so they could be used by other platforms that use dual stick controllers instead of just ifdefing code that hard codes everything to one platform

Defaulting the player name to "The Stranger" is not appropriate as it is not something the original games did and at this point in the timeline the Avatar was no longer known as The Stranger. Seems like this is something of a personal preference put into the code.

Changes to rip and wuc, why? These are commandline tools and shouldn't even be built for non computer systems if they are there probably needs to be changes to the build system so they are only built when appropriate.

einmueller:
All critic of wench is accepted. Then just close this and for now (and for  my spare time) I'll do the vita port as separate project and see how much work is needed to merge your fixes. Perhaps later with one huge patch request, instead of so much little ones. There are not so many functional changes these days, so it'll be ok, I think. Most Vita-Users will only use the binaries, so the source would only be useful for one or two gods who perhaps could really improve things. They'll use my repo then.

@rofl0r: These are my very first steps with git. It's a kind of hippie thing for me. I'm an old school hobby programmer, grown with vi, diff and patch. I even never programmed with c++, only with plain c and SDL coding was very new to me. So yes, many commit because git is a beast. And yes perhaps not the best style, I know. You should really learn to be more kindly with people who want to contribute to exult and help getting things better, even if they seem to be idiots for you. We all do this as a hobby!

PR #232, opened 2022-08-19
Mandatory Pkg-Config in Linux and MacOS builds for Exult 1.9+

Dragon-Baroque:
Commit 1 of 1 :
    configure.ac
      After Pkg-Config detection, fail if not found
      Remove tests over Pkg-Config existence, keep then branch, remove else branch
      Comment out SYSLIBS set to X11 when building with Exult Studio support

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -270,6 +270,9 @@ fi
 # Check for pkg-config
 # ---------------------------------------------------------------------
 PKG_PROG_PKG_CONFIG
+if test x$PKG_CONFIG = x; then
+	AC_MSG_ERROR([*** To Build Exult 1.9+, pkg-config is required. ***])
+fi
 
 # ---------------------------------------------------------------------
 # SDL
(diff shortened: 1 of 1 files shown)


flamewing (approved):

marzojr (approved):
It looks a lot nicer. I just want to wait for the LGTM analysis before merging; they use a strange build system autodetection stuff that can error out (though I believe it will not in this case).

Dragon-Baroque:
> It looks a lot nicer. I just want to wait for the LGTM analysis before merging; they use a strange build system autodetection stuff that can error out (though I believe it will not in this case).

Plus the correct completion of the MacOS build that I cannot perform myself.

marzojr:
Seems only the Discord notification is failing.

Hm. Seems [LGTM is shutting down](https://github.blog/2022-08-15-the-next-step-for-lgtm-com-github-code-scanning/), because it has been folded into GH Code Scanning. Guess I should investigate how to go about moving to it.

PR #234, opened 2022-08-19
Silent Clang++ compiler over SDL

Dragon-Baroque:
Commit 1 of 1 :
    .cc and .h
      Bracket includes of SDL by -Wold-style-cast
      Create static for SDL_TOUCH_MOUSEID

### DIFF
--- a/audio/Audio.h
+++ b/audio/Audio.h
@@ -22,8 +22,17 @@
 #include <map>
 #include <memory>
 #include <vector>
+
+#ifdef __GNUC__
+#	pragma GCC diagnostic push
+#	pragma GCC diagnostic ignored "-Wold-style-cast"
+#endif    // __GNUC__
 #include <SDL.h>
 #include <SDL_audio.h>
+#ifdef __GNUC__
+#	pragma GCC diagnostic pop
+#endif    // __GNUC__
+
 #include "AudioMixer.h"
 #include "exceptions.h"
 #include "exult_constants.h"
--- a/audio/AudioMixer.cc
+++ b/audio/AudioMixer.cc
@@ -24,7 +24,14 @@ Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.
 
 #include "MidiDriver.h"
 
+#ifdef __GNUC__
+#	pragma GCC diagnostic push
+#	pragma GCC diagnostic ignored "-Wold-style-cast"
+#endif    // __GNUC__
 #include <SDL.h>
+#ifdef __GNUC__
+#	pragma GCC diagnostic pop
+#endif    // __GNUC__
 
 #include <algorithm>
 #include <iostream>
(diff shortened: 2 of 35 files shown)


rofl0r:
seems quite ugly to hardcode this into every TU.
why not check in configure whether -Wno-old-style-cast is accepted by the compiler, then pass it as CXXFLAGS to any file compiled ?

Dragon-Baroque:
> Why not check in configure whether -Wno-old-style-cast is accepted by the compiler, then pass it as CXXFLAGS to any file compiled ?

It is actually the opposite. We want to be warned of any old style cast ( that is any C style cast ) remaining in the Exult C++ code, but there are old style casts in included tools : SDL ( Exult ) , GTK+ ( Exult Studio ).

Your point is valid though, the SDL includes are not well organized, compared to the GTK+ includes of Studio which are factored into `studio.h`. Time for some more study, I guess. I suspect that the lack of a common include of `SDL.h` for all of Exult is that some of the Exult code is shared with Studio and that SDL and GTK+ would badly coexist in Studio.

I think we should
- Switch the SDL includes to angle brakets instead of double quotes. Easy, I can do that.
- Avoid the replicas, for example in `txtscroll.cc` the includes of SDL_stdinc and SDL_scancode are already done by SDL_events. Actually, where  `SDL_events.h` is included, there is seldom a need for some other SDL include.
- Understand why we cannot include the whole of `SDL.h` whenever SDL is needed. Is this because `SDL.h` drags the `SDL_main.h`, which causes problems with Android ?
- Apart from the independent `SDL_syswm.h` of SDL2 for detailed system events, and from `SDL_image.h` of SDL2_image for the tools smooth and mockup ( are they alive ? ), we could split the SDL2 `SDL.h` include into `SDL_events.h` ( input + video ) and `SDL_audio.h` ( audio ) and discard all other `SDL_xxx.h` includes.

What do you think @DominusExult @marzojr @ceckak ?

DominusExult:
I'm all for streamlining everything. But I have no idea why not just pull in SDL.h everytime. You might be right with the SDL_main.h being pulled in then.

I think the reason why the Discord notification fails on pull requests is that the CI is done on the pull requesters instance and not Exult's. So the Exult webhook secret is/can not be used.

ceckak:
I am also fully supportive of using SDL.h everywhere.  I don't expect it will cause any android problems - it's fine if SDL_main.h gets pulled in as that just ensures we don't have a real main function.  The only exception to SDL.h replacing other SDL includes are if there are some that it *doesn't* pull in that are needed for platform-specifics.  I'm happy to backfill fixes for anything that breaks on Android.  I'm overdue for a rebase and submitting some additional Android fixes.

I'm also in favor of using angle brackets for all includes other than cases where you want to include a header from the same directory as the source file.

marzojr (approved):

PR #246, opened 2022-09-03
Add FreeBSD platform

nunotexbsd:


nunotexbsd (commented):
Add FreeBSD platform

marzojr (approved):

nunotexbsd:
Thanks @marzojr

nunotexbsd:
Sorry about my inexperience but I think I missed a tab in version.cc:
```
#elif (defined(FREEBSD))
out << "FreeBSD";
#elif (defined(MACOSX))
	out << "Mac OS X";
```

DominusExult:
I wonder if we should replace this with fewer ifdefs with SDl_GetPlatform https://wiki.libsdl.org/SDL_GetPlatform
It's pretty generic but apart from Windows version detection we were just as basic anyway.

wench:
Yep SDL_GetPlatform seems to exist exactly for this reason. Shame it doesn't give more detailed version information like we do for Windows. Seems like we should use it

PR #248, opened 2022-09-12
Set all includes of SDL into angle brackets not double quotes, promote to SDL.h

Dragon-Baroque:
- Prefer `#include <SDLblabla.h>` to `#include "SDLblabla.h"`. Apply everywhere.
- Prefer `#include <SDL.h>` to `#include <SDLblabla.h>`. Apply everywhere except for `tools/shp2pcx.cc` with remains with `#include <SDL_endian.h>` because it only uses the `SDL_BYTEORDER` endianness macro at compile time, but not any SDL service at run time. Moreover, including full `SDL.h` in a stand alone executable, which `shp2pcx` is, disturbs the `main()` logic, as it is renamed into `SDL_main()`.

### DIFF
--- a/gumps/Gump_manager.cc
+++ b/gumps/Gump_manager.cc
@@ -22,22 +22,20 @@
 #  include <config.h>
 #endif
 
-#ifdef __GNUC__
-#	pragma GCC diagnostic push
-#	pragma GCC diagnostic ignored "-Wold-style-cast"
-#endif    // __GNUC__
-#include "SDL_events.h"
-#include "SDL_keyboard.h"
-static const Uint32 EXSDL_TOUCH_MOUSEID=SDL_TOUCH_MOUSEID;
-#ifdef __GNUC__
-#	pragma GCC diagnostic pop
-#endif    // __GNUC__
-
 #include "Configuration.h"
 #include "exult.h"
 #include "Gump.h"
 #include "Gump_manager.h"
 #include "gamewin.h"
+
+#ifdef __GNUC__
+#	pragma GCC diagnostic push
+#	pragma GCC diagnostic ignored "-Wold-style-cast"
+#endif    // __GNUC__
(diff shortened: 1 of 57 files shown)


gumps/Gump_manager.cc
```diff
@@ -22,23 +22,21 @@
...
 #ifdef __GNUC__
 #	pragma GCC diagnostic push
 #	pragma GCC diagnostic ignored "-Wold-style-cast"
 #endif    // __GNUC__
-#include "SDL_events.h"
-#include "SDL_keyboard.h"
 static const Uint32 EXSDL_TOUCH_MOUSEID=SDL_TOUCH_MOUSEID;
 #ifdef __GNUC__
 #	pragma GCC diagnostic pop
 #endif    // __GNUC__
```
marzojr, line 38: Shouldn't the pragmas and ifdefs be removed as well?

marzojr (changes requested):

gumps/Gump_manager.cc
Dragon-Baroque: I am afraid not because `SDL_TOUCH_MOUSEID` is a macro with a C style cast in it :
```
#define SDL_TOUCH_MOUSEID ((Uint32)-1)
```
The whole point of storing into `EXSDL_TOUCH_MOUSEID` is to hide this cast in a safe location in a `-Wno-old-style-cast` frame  and use `EXSDL_TOUCH_MOUSEID` everywhere else.

rofl0r:
i personally prefer `<SDL2/SDL.h>` because this allows to use standard header search paths of the compiler, whereas just `<SDL.h>` requires to add a full include path to the compiler command line, which is usually retrieved via the non-cross-compile-compatible `sdl2config` tool.

that way a configure script checking for sdl2 can just compile a small file doing `#include <SDL2/SDL.h>     int foo;` and check the return value, which should be success if the headers are installed in the default search path.

marzojr (approved):

Issue #249, opened 2022-09-16
Exult Studio: Files/Groups list need to be scrollable


DominusExult:
Exult Studio may become hard to use because the files/groups list can't be made smaller than the list needs to be when every file type is uncollapsed.
Example: 
- Open a Serpent Isle game in Exult Studio
- now try to resize the height of the Exult Studio window
- It will not get smaller than the last file in the list (unless you collapse Palette Files)

This is a problem on smaller laptop screens (1366x768) or when you have a lot of extra files in Shape Files, Palette Files, ...
The person who reported it was desperate because he couldn't resize the window and thus not use the buttons on the bottom.
Only after trying a lot of things  did we realize that this was the culprit.

@Dragon-Baroque could you take a look? I guess you are the most familiar at the moment.

![image](https://user-images.githubusercontent.com/3313896/190685459-651080f9-c642-4271-939a-a6f409cb941f.png)

Dragon-Baroque:
@DominusExult, there is a very simple change that I would like you to perform and tell me whether it suits you.
Change with a regular text editor in the `share/exult/exult_studio.glade` the Vertical Scroll Policy of the `scrolledwindow2` from `never` to `always` :
```
...
              <object class="GtkScrolledWindow" id="scrolledwindow2">
                <property name="visible">True</property>
                <property name="can-focus">False</property>
                <property name="halign">fill</property>
                <property name="valign">fill</property>
                <property name="margin-start">4</property>
                <property name="margin-end">4</property>
                <property name="margin-top">2</property>
                <property name="margin-bottom">2</property>
                <property name="hexpand">False</property>
                <property name="vexpand">False</property>
                <property name="hscrollbar-policy">never</property>
                <property name="vscrollbar-policy">never</property>
...
```
You should alter the last line of that excerpt to
```
                <property name="vscrollbar-policy">always</property>·
```
Then run Exult Studio again.
This, for me, creates a vertical scroll bar on the file / group treeview, visible when I hover the mouse over its expected place at the right of the treeview.
But it makes also the treeview _very, very small_ vertically, I mean only five rows or so are visible.
Does it suit you ?
If so I can push a commit to patch the glade file

DominusExult:
Works fine, also works with setting it to automatic. I also see the smallness of the vertical treeview, can there be a default expansion? Or maybe a default for the whole Studio windows (as it is saved on closing in exult.cfg)?
[...]

PR #253, opened 2022-09-28
Support mouse wheel vertical scrolling in shape browsers

Dragon-Baroque:
To complete issue #249 :  

- Declare a ( static ) Scroll Event Handler in `mapedit/chunklst.h + combo.h + npclst.h + shapelst.h`
- Create and Bind a Scroll Event Handler in the browser constructor in ` mapedit/chunklst.cc + combo.cc + npclst.cc + shapelst.cc`
- Implement the Scroll Event handler in ` mapedit/chunklst.cc + combo.cc + npclst.cc + shapelst.cc` emulating the GTK effect of a mouse wheel event _on the vertical scroller_ :

1. Mouse wheel Scroll Events _on non macOS_ are in click units, then send a Scroller move of
`numberOfClicks * ( browserHeight ** 2/3 )`,
3. Mouse wheel Scroll Events _on macOS_ are in pixel units, then send a Scroller move of `numberOfPixels`.

### DIFF
--- a/mapedit/chunklst.cc
+++ b/mapedit/chunklst.cc
@@ -728,6 +728,8 @@ Chunk_chooser::Chunk_chooser(
 	g_signal_connect(G_OBJECT(chunk_adj), "value-changed",
 	                 G_CALLBACK(scrolled), this);
 	gtk_widget_show(vscroll);
+	// Scroll events.
+	enable_draw_vscroll(draw);
 
 	// At the bottom, status bar:
 	GtkWidget *hbox1 = gtk_box_new(GTK_ORIENTATION_HORIZONTAL, 0);
--- a/mapedit/combo.cc
+++ b/mapedit/combo.cc
@@ -1161,6 +1161,8 @@ Combo_chooser::Combo_chooser(
 	g_signal_connect(G_OBJECT(combo_adj), "value-changed",
 	                 G_CALLBACK(scrolled), this);
 	gtk_widget_show(vscroll);
+	// Scroll events.
+	enable_draw_vscroll(draw);
 
 	// At the bottom, status bar:
 	GtkWidget *hbox1 = gtk_box_new(GTK_ORIENTATION_HORIZONTAL, 0);
--- a/mapedit/npclst.cc
+++ b/mapedit/npclst.cc
@@ -1083,6 +1083,8 @@ Npc_chooser::Npc_chooser(
 	g_signal_connect(G_OBJECT(shape_adj), "value-changed",
 	                 G_CALLBACK(hscrolled), this);
 //++++  gtk_widget_hide(hscroll);   // Only shown in 'frames' mode.
+	// Scroll events.
+	enable_draw_vscroll(draw);
 
 	// At the bottom, status bar & frame:
 	GtkWidget *hbox1 = gtk_box_new(GTK_ORIENTATION_HORIZONTAL, 0);
--- a/mapedit/shapelst.cc
+++ b/mapedit/shapelst.cc
@@ -2314,6 +2314,8 @@ Shape_chooser::Shape_chooser(
 	g_signal_connect(G_OBJECT(shape_adj), "value-changed",
 	                 G_CALLBACK(hscrolled), this);
 //++++  gtk_widget_hide(hscroll);   // Only shown in 'frames' mode.
+	// Scroll events.
+	enable_draw_vscroll(draw);
 
 	// At the bottom, status bar & frame:
 	GtkWidget *hbox1 = gtk_box_new(GTK_ORIENTATION_HORIZONTAL, 0);
--- a/mapedit/gtk_redefines.h
+++ b/mapedit/gtk_redefines.h
@@ -48,10 +48,12 @@ inline T1 *gtk_cast(T2 *obj) {
 #  undef _G_TYPE_CIC
 #  undef _G_TYPE_CCC
 #  undef _G_TYPE_CIT
+#  undef _G_TYPE_CIFT
 #  ifndef G_DISABLE_CAST_CHECKS
 #    define _G_TYPE_CIC(ip, gt, ct)     gtk_cast<ct>(g_type_check_instance_cast(gtk_cast<GTypeInstance>((ip)), gt))
 #    define _G_TYPE_CCC(cp, gt, ct)     gtk_cast<ct>(g_type_check_class_cast(gtk_cast<GTypeClass>((cp)), gt))
 #    define _G_TYPE_CIT(ip, gt)                     (g_type_check_instance_is_a(gtk_cast<GTypeInstance>((ip)), gt))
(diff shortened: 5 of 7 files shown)


mapedit/chunklst.cc
```diff
@@ -728,6 +753,12 @@ Chunk_chooser::Chunk_chooser(
 	g_signal_connect(G_OBJECT(chunk_adj), "value-changed",
 	                 G_CALLBACK(scrolled), this);
 	gtk_widget_show(vscroll);
+	// Scroll events.
+	GtkEventController *scroll_ctlr = GTK_EVENT_CONTROLLER(
```
marzojr, line 757: Hm. I wonder if scroll_ctlr and hbox1 below need to be added to the class and be destroyed when the Chunk_chooser is deleted. Most of the others are handled by gtk_box_pack_start or set_widget, but I wonder if they also need something like a call to g_object_unref to avoid leaking.

mapedit/combo.cc
```diff
@@ -1161,6 +1187,12 @@ Combo_chooser::Combo_chooser(
 	g_signal_connect(G_OBJECT(combo_adj), "value-changed",
 	                 G_CALLBACK(scrolled), this);
 	gtk_widget_show(vscroll);
+	// Scroll events.
+	GtkEventController *scroll_ctlr = GTK_EVENT_CONTROLLER(
```
marzojr, line 1191: Same here regarding unrefs.

mapedit/chunklst.cc
```diff
@@ -566,6 +567,30 @@ void Chunk_chooser::scrolled(
...
+	auto *chooser = static_cast<Chunk_chooser *>(data);
+	GtkAdjustment *adj = gtk_range_get_adjustment(
+	    GTK_RANGE(chooser->vscroll));
+	const gdouble adj_value  = gtk_adjustment_get_value(adj);
+#if defined(MACOSX) && !defined(XWIN)
+	const gdouble new_unit   = 1.0;
+#else
+	const gdouble adj_pgsize = gtk_adjustment_get_page_size(adj);
+	const gdouble new_unit   = pow(adj_pgsize, 2.0/3.0);
+#endif // MACOSX && !XWIN
```
marzojr, line 589: Would you mind extracting these lines into a function so that the #ifdef hell can be confined to only one place and call on all the scroll functions? See other notes.

mapedit/combo.cc
```diff
@@ -1039,6 +1041,30 @@ void Combo_chooser::scrolled(
...
+	auto *chooser = static_cast<Combo_chooser *>(data);
+	GtkAdjustment *adj = gtk_range_get_adjustment(
+	    GTK_RANGE(chooser->vscroll));
+	const gdouble adj_value  = gtk_adjustment_get_value(adj);
+#if defined(MACOSX) && !defined(XWIN)
+	const gdouble new_unit   = 1.0;
+#else
+	const gdouble adj_pgsize = gtk_adjustment_get_page_size(adj);
+	const gdouble new_unit   = pow(adj_pgsize, 2.0/3.0);
+#endif // MACOSX && !XWIN
```
marzojr, line 1063: Extract to common function?

mapedit/npclst.cc
```diff
@@ -846,6 +846,30 @@ void Npc_chooser::hscrolled(      // For horizontal scrollbar.
...
+	auto *chooser = static_cast<Npc_chooser *>(data);
+	GtkAdjustment *adj = gtk_range_get_adjustment(
+	    GTK_RANGE(chooser->vscroll));
+	const gdouble adj_value  = gtk_adjustment_get_value(adj);
+#if defined(MACOSX) && !defined(XWIN)
+	const gdouble new_unit   = 1.0;
+#else
+	const gdouble adj_pgsize = gtk_adjustment_get_page_size(adj);
+	const gdouble new_unit   = pow(adj_pgsize, 2.0/3.0);
+#endif // MACOSX && !XWIN
```
marzojr, line 868: Extract to common function?

mapedit/npclst.cc
```diff
@@ -1084,6 +1108,13 @@ Npc_chooser::Npc_chooser(
 	                 G_CALLBACK(hscrolled), this);
 //++++  gtk_widget_hide(hscroll);   // Only shown in 'frames' mode.
 
+	// Scroll events.
+	GtkEventController *scroll_ctlr = GTK_EVENT_CONTROLLER(
```
marzojr, line 1112: Same here regarding unrefs.

mapedit/shapelst.cc
```diff
@@ -2315,6 +2339,13 @@ Shape_chooser::Shape_chooser(
 	                 G_CALLBACK(hscrolled), this);
 //++++  gtk_widget_hide(hscroll);   // Only shown in 'frames' mode.
 
+	// Scroll events.
+	GtkEventController *scroll_ctlr = GTK_EVENT_CONTROLLER(
```
marzojr, line 2343: Same here regarding unrefs.

mapedit/shapelst.cc
```diff
@@ -1863,6 +1863,30 @@ void Shape_chooser::hscrolled(      // For horizontal scrollbar.
...
+	auto *chooser = static_cast<Shape_chooser *>(data);
+	GtkAdjustment *adj = gtk_range_get_adjustment(
+	    GTK_RANGE(chooser->vscroll));
+	const gdouble adj_value  = gtk_adjustment_get_value(adj);
+#if defined(MACOSX) && !defined(XWIN)
+	const gdouble new_unit   = 1.0;
+#else
+	const gdouble adj_pgsize = gtk_adjustment_get_page_size(adj);
+	const gdouble new_unit   = pow(adj_pgsize, 2.0/3.0);
+#endif // MACOSX && !XWIN
```
marzojr, line 1885: Extract to common function?

marzojr (changes requested):

mapedit/chunklst.cc
Dragon-Baroque: You are right about `scroll_ctlr` needing a `g_object_unref`. It is not a `GtkWidget` and it is not created `floating`.
I have fully factorized the scroller in the browsers into the class `Object_browser` which is the parent class to all browsers, and I have put the `g_object_unref` into `Object_browser`'s destructor.

On the other hand, I do not see why you single `hbox1`, as it is also `gtk_box_pack_started` into `vbox`. I have written a quick check as putting it into `Object_browser` and checking its state in `Object_browser`'s destructor : it is destroyed when the destructor is invoked.

mapedit/chunklst.cc
Dragon-Baroque: I could do much more. Factorizing everything into the parent class `Object_browser` ( `objbrowse.cc` and `objbrowse.h` ). There remains only the activation of the browser scroll from each daughter class ( minus the Palette browser ), since one need he `GtkDrawingArea` and also since it cannot be set in the `Object_browser` constructor, much too early.

marzojr (approved):

Issue #283, opened 2022-12-01
Defects in Exult and Studio for Chunks and Terrain


Dragon-Baroque:
**Chunk display defect in Studio**

Studio does not display correctly some Chunks, they differ from Exult ( Taken from Issue #278 ) :

> ![Untitled](https://user-images.githubusercontent.com/74261498/204656718-2190e808-0243-41f5-91ea-e8cd8639eb92.png)

> This snapshot is from Studio at master, over Forge of Virtue, no mod.
The Chunk at about the center, with the mountain on the top left quarter and the coast on the three other quarters is number 65, and it displays a strange greenish horizontal line at the vertical middle, half right part.
Chunk 66 also shows a brownish horizontal line at about the same place, plus the bottom border looks like one pixel down with the right quarter.
Chunk 64 has three black inverted L and so has the Chunk above it, one of which at the bottom looks fully shifted one pixel down and right.

In Exult's `gamerend.cc`, `Game_render::paint_map` makes _two_ passes on the chunk, then adds the objects.

- Each pass processes the 16 * 16 shapes of the chunk,
- The first pass paints the `non RLE` shapes, those are 8*8 tiles with a `shapenum` less than 150; the shapes are not clickable, nor hack-movable; their X and Y origin is defaulted to -1 which is wrong and shifts them in the display,
- The second pass paints the `RLE` shapes, those are true flat objects larger than 8*8 pixels with a `shapenum` larger or equal than 150; the shapes are clickable and then they display their name ( and their outline in cheat mode ); they are also hack-movable, but then they leave a black square; finally their X and Y origin is correct.

I have altered `render_chunk` in `mapedit/chunklst.cc` to implement the two passes of Exult and the adjustment of the origin, and the display of the Chunks in Studio now looks correct. I can post the patch or make a Pull Request. Please let me know.

Dragon-Baroque:
**Display defect and Crash in Exult when Studio's Edit Terrain is enabled**

When `Edit Terrain` is enabled from Studio, Exult display the Chunks/Terrains using an accelerated path `Game_render::paint_terrain_only` in `gamerend.cc` into `Chunk_terrain::render_all` in `objs/chunkter.cc`. 
This pair of functions has two defects :
[...]

PR #291, opened 2022-12-09
Fix enabling Edit Terrain after Locating a chunk near a superchunk boundary

Dragon-Baroque:
Fixes Issue #283 :

Discrepancy in the chunk cache between its build in `Game_map::read_map_data` and its use in `Game_render::paint_map` when the Edit Terrain mode is enabled.
The Smooth Scrolling mode in commit 237dc6881 requires one more chunk at the right and at the bottom in `Game_render::paint_map` but the chunk cache in `Game_map::read_map_data` was not altered.

This PR increases the chunk limits when building the chunk cache.


marzojr (approved):
Looks good to me. I assume this causes no other crashes?

Dragon-Baroque:
>  I assume this causes no other crashes?

I have not got any other crash. I am quite ready to debug any that could come, though.

PR #293, opened 2022-12-11
Fix crash in Cheat menu when holding Shift for some time

Dragon-Baroque:
When expecting a Name, a Decimal number or an Hexadecimal number, the Cheat menu crashes when the user holds a control key ( Shift, Control, ... ) for some time.

Then SDL sends a key event, whose SDL_Keysym is outside the range 0..255 and then `std::isdigit`, `std::isxdigit`, `std::isalnum` index a `ctype` array outside its bounds.
Ensure that the SDL_Keysym is in the range 0..255 or return false to the check.

### DIFF
--- a/cheat_screen.cc
+++ b/cheat_screen.cc
@@ -477,7 +477,7 @@ bool CheatScreen::SharedInput(char *input, int len, int &command, Cheat_Prompt &
 					activate = true;
 				} else if ((key.sym == '-' || key.sym == SDLK_KP_MINUS) && !input[0]) {
 					input[0] = '-';
-				} else if (std::isxdigit(key.sym)) {
+				} else if ((key.sym < 256 && key.sym >= 0 ? std::isxdigit(key.sym) : false)) {
 					const int curlen = std::strlen(input);
 					if (curlen < (len - 1)) {
 						input[curlen] = std::tolower(key.sym);
@@ -497,7 +497,7 @@ bool CheatScreen::SharedInput(char *input, int len, int &command, Cheat_Prompt &
 			} else if (mode >= CP_Name) {      // Want Text input (len chars)
 				if (key.sym == SDLK_RETURN || key.sym == SDLK_KP_ENTER) {
 					activate = true;
-				} else if (std::isalnum(key.sym) || key.sym == ' ') {
+				} else if ((key.sym < 256 && key.sym >= 0 ? std::isalnum(key.sym) : false) || key.sym == ' ') {
 					const int curlen = std::strlen(input);
 					char chr = key.sym;
 					if (key.mod & KMOD_SHIFT) {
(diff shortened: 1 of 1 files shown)


cheat_screen.cc
```diff
@@ -477,7 +477,7 @@ bool CheatScreen::SharedInput(char *input, int len, int &command, Cheat_Prompt &
 					activate = true;
 				} else if ((key.sym == '-' || key.sym == SDLK_KP_MINUS) && !input[0]) {
 					input[0] = '-';
-				} else if (std::isxdigit(key.sym)) {
+				} else if ((key.sym < 256 && key.sym >= 0 ? std::isxdigit(key.sym) : false)) {
```
rofl0r, line 480: seems like the right fix would be std::isxdigit((unsigned)key.sym)

see also https://drewdevault.com/2020/09/25/A-story-of-two-libcs.html for why this crashes at all.

wench: cast to unsigned Maybe I really couldn't tell as this is a strange platform specific problem. i'd have though cast to unsigned char would be better
but what I do know is I do not like the originial fix.
Use of the ternary operator in an if statement is all types of ick especially when && short circuiting will be sufficient

Dragon-Baroque: No sorry, cast would not do. Admitted, a `SDL_Keysym` like `key . sym` is a signed int 32.  But its value is either an ASCII code ( when possible ), or a Scancode marker 0x4000000 ored with the Scancode.
Casting to `unsigned` would in fact do nothing, and casting to `unsigned char` would risk false positives when the Scancode [ please mind : NLS dependent ] collides with a regular 0..9 a..f ASCII.

Thus I need the range checks, and you are right, it it amazing that `std::isdigit` and al expect a `int` argument but index a 256 elements array.

You are right too, the ternary has no use at this point, I cannot figure why I came to use it : I am submitting a replacement shortly.

marzojr: For what is worth, the intention behind those functions was that you could pass in the result of fgetc and similar functions, which return either EOF (`(int)-1`) or a value that can be stored in an unsigned char. They do handle an incoming EOF well, but when you factor in that char is signed in some platforms and people will be using these functions for string processing, it ends up making for a terrible design overall.

Dragon-Baroque: > For what is worth, the intention behind those functions was that you could pass in the result of fgetc and similar functions, which return either EOF ((int)-1) or a value that can be stored in an unsigned char.

Aha. I thought that the `int` signature would be for Unicode instead. 

I wondered about discarding the `std::is...` entirely and rewrite that code using only the `SDLK_..` macros, even when handling regular ASCII ( `SDLK_n` is `'n'` ).

Since there are no `SDLK_N` and similar, only `SDLK_n` with a `KMOD_SHIFT` modifier, the range checks equivalent to `isdigit`, `isxdigit` and `isalnum` are easier, there is no need to check for lower case _and_ uppercase. In fact I discovered later that I could key in the regular decimal digits _without_ the Shift key, as I need to normally. Still, when I key in 'a' I get 'a' not 'z'. Thus SDL does some kind of NLS normalization that I cannot figure out.

Removing the `std::is...` would make the cheat code marginally more verbose. The problem of the numeric keypad `SDLK_KP_0` coming after contiguous `SDLK_KP_1..9` has been already handled.

marzojr: > Aha. I thought that the `int` signature would be for Unicode instead.
That would have been better. Unfortunately, Unicode was still a few years in the future when these functions were created.

As is, c and c++ specify it is undefined behavior to use those functions when the input is not one of those values I mentioned (EOF or fit in unsigned char), making those functions trivially easy to misuse.

PR #322, opened 2023-03-16
SIfixes add face for SmithzHorse, supporting files

KnightCaptainU7:
Adds the graphics\faces folder for the face for SmithzHorse as created by Rhona. Number is padded to match my (Knight Captain's) mod.
Fixes the conversation for the nightmare horse shape to use that face, conversation is also split up to support different video resolutions. Note that the NPC has 0 health to start with if you test with a brand new game and hack-move. Tested the nightmare's death in the Mountains of Freedom.
Also adds one utility usecode in full, which could be used by other modders in original Exult games.


content/sifixes/src/header/si/si_npcs.uc
```diff
@@ -328,7 +328,8 @@ enum northernforest_npcs
 	IVOR = -158,
 	MORGHRIM = -159,
 	HOUND_OF_DOSKAR = -269,
-	FACE_HOUND_OF_DOSKAR = -279,
+	// Moved to si_faces.uc
+	// FACE_HOUND_OF_DOSKAR = -279,
```
marzojr, line 332: Better to remove these lines.

content/sifixes/src/header/si/si_faces.uc
```diff
@@ -0,0 +1,223 @@
...
+	FACE_HAZARD = 280,
+	FACE_ERNESTO = 281,
+	FACE_ANTI_DUPRE1 = 282,
+	// 283 unused
+	FACE_CHAOS_HIEROPHANT = 284,
+	// 285 unused
+	FACE_BEATRIX = 286,
+	FACE_BATLIN = 287,
+		FRAME_WITH_MEDALLION = 0,
+		FRAME_WITHOUT_MEDALLION = 1,
```
marzojr, line 193: I think that the FRAME_XXX constants should be renamed to FRAME_<NPCNAME>_XXX. For example, `FRAME_COLUMNA_YOUNG` instead of `FRAME_YOUNG`.

marzojr: Also, maybe move them to a separate enum and remove the double indent.

marzojr (approved):

Issue #334, opened 2023-04-01
RFC: Unicode in UCC/UCXT


marzojr:
One thing I was thinking to make usecode easier to read/write (particularly in the French and German translations) is making UCXT dump the text as unicode (maybe as an option), and UCC translate these unicode characters back into the native character maps. Maybe also adding a way to specify a character set (like ES does) if you want an official ANSI code page instead.

The practical effect would be that UCXT would (could?) output this:
```"@Die Segel müssen eingeholt werden, bevor die Planken\r\n\t\t\terhöht werden können.@"```

instead of this:
```"@Die Segel mssen eingeholt werden, bevor die Planken\r\n\t\t\terhht werden knnen.@"```

and UCC would handle the reverse conversion. For what is worth, I am thinking of using a similar scheme for exultmsg and textmsg files so that they are easier to read/edit.

But there is one additional wrinkle I thought: we could use a scheme for runic alphabet in unicode (e.g., [this one](https://www.reddit.com/r/Ultima/comments/p1rf4x/runes_in_unicode/) and have this:
```"·ᛳᚫᚳᛷ·", "ᛠᛏᛖᚾ", "ᛒᚥ", "ᛞᛁᛟᚩᛋ"```

instead of this:
```"|jack|", "+ten", "by", "di*os"```

This runic text conversion would be difficult to do in UCXT automatically, though, but UCC could handle it without issues.

Needless to say, the old way of doing this would still work. Any thoughts?

DominusExult:
I'm tagging @ungesundes-halbwissen  as he might have some ideas.
But to me it sounds like a great idea

Dragon-Baroque:
Yes, I would like UCC / UCXT to support Unicode.
You could consider adding an option to UCXT, but not to UCC, if, say, UCXT created the Unicode decompilation with a BOM ( Byte Order Mark ) at the beginning of the file. Then UCC could check the source file for a BOM at the beginning and use the file as Unicode if present, ANSI if missing, like Windows does.
The BOM is U+FEFF [ Zero Width NoBreak Space ], its UTF-8 encoding is EF BB BF.

marzojr:
[...]

PR #337, opened 2023-04-05
Support Ultima VII German / French 7 bit ASCII codepage with reuse of control bytes for diacritics

Dragon-Baroque:
Commit 1 of 1 :
    gamemgr/modmgr.h + cc, mapedit/studio.cc
      Set default codepage to ASCII : 7 bits with reuse
    mapedit/exult_studio.glade, mapedit/studio.cc
      Add codepages ASCII 7 bits and CP-850 into Game Information -> Character Set combo box
    mapedit/shapelst.cc
      Add missing conversions to UTF8 in update_statusbar
    mapedit/studio.cc
      Display half size encoding table when using ASCII 7 bits
      Display byte 08, produces binary 08 unless reused by ASCII
      Conversion from ASCII to UTF8 : front the conversion by a move of diacritics to PC437 location
      Conversion from UTF8 to ASCII : back  the conversion by a move of diacritics to control bytes

This is a proposal related to the issue #334.

### DIFF
--- a/mapedit/studio.cc
+++ b/mapedit/studio.cc
@@ -1214,7 +1214,7 @@ C_EXPORT void on_gameselect_ok_clicked(
 		GtkTextIter endpos;
 		gtk_text_buffer_get_bounds(buff, &startpos, &endpos);
 		gchar *modmenu = gtk_text_iter_get_text(&startpos, &endpos);
-		const codepageStr menu(modmenu, "CP437");
+		const codepageStr menu(modmenu, "ASCII");
 		const string modmenustr = menu.get_str();
 		g_free(modmenu);
 		if (modmenustr.empty())
@@ -1313,7 +1313,7 @@ C_EXPORT void on_gameselect_gamelist_cursor_changed(
 
 		// Titles need to be displayable in Exult menu, hence should not
 		// have any extra characters.
-		const utf8Str title(modname.c_str(), "CP437");
+		const utf8Str title(modname.c_str(), "ASCII");
 		gtk_tree_store_append(model, &iter, nullptr);
 		gtk_tree_store_set(model, &iter,
 		                   0, title.get_str(),
@@ -1338,7 +1338,7 @@ void fill_game_tree(GtkTreeView *treeview, int curr_game) {
 
 		// Titles need to be displayable in Exult menu, hence should not
 		// have any extra characters.
-		const utf8Str title(gamename.c_str(), "CP437");
+		const utf8Str title(gamename.c_str(), "ASCII");
 		gtk_tree_store_append(model, &iter, nullptr);
 		gtk_tree_store_set(model, &iter,
 		                   0, title.get_str(),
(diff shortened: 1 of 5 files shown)


marzojr (approved):

mapedit/studio.cc
```diff
@@ -3190,8 +3208,45 @@ void convertFromUTF8::convert(gchar *&_convstr, const char *str, const char *enc
...
+				else if (*ch <= 0x9f) ;
+				// a0 -> 19 á:a acute.
+				else if (*ch <= 0xa0) *ch -= 0x87;
+				// a1 -> 1d í:i acute,      a2 -> 1e ó:o acute,      a3 -> 1f ú:u acute.
+				else if (*ch <= 0xa3) *ch -= 0x84;
+				else if (*ch <= 0xe0) ;
+				// e1 -> 1c ß:s sharp.
+				else if (*ch <= 0xe1) *ch -= 0xc5;
+			}
+		}
```
marzojr, line 3247: This whole block of code appears twice; it probably should be extracted to a function.

mapedit/studio.cc
```diff
@@ -3225,21 +3280,57 @@ void convertFromUTF8::convert(gchar *&_convstr, const char *str, const char *enc
...
+			else if (*ch <= 0x9f) ;
+			// a0 -> 19 á:a acute.
+			else if (*ch <= 0xa0) *ch -= 0x87;
+			// a1 -> 1d í:i acute,      a2 -> 1e ó:o acute,      a3 -> 1f ú:u acute.
+			else if (*ch <= 0xa3) *ch -= 0x84;
+			else if (*ch <= 0xe0) ;
+			// e1 -> 1c ß:s sharp.
+			else if (*ch <= 0xe1) *ch -= 0xc5;
+		}
+	}
```
marzojr, line 3333: This is the second copy.

mapedit/studio.cc
```diff
@@ -3252,6 +3343,46 @@ void convertToUTF8::convert(gchar *&_convstr, const char *str, const char *enc)
...
+				// 99 <- 17 Ö:O diaeresis,  9a <- 18 Ü:U diaeresis.
+				else if (*ch <= 0x18) *ch += 0x82;
+				// a0 <- 19 á:a acute.
+				else if (*ch <= 0x19) *ch += 0x87;
+				else if (*ch <= 0x1b) ;
+				// e1 <- 1c ß:s sharp.
+				else if (*ch <= 0x1c) *ch += 0xc5;
+				// a1 <- 1d í:i acute,      a2 <- 1e ó:o acute,      a3 <- 1f ú:u acute.
+				else if (*ch <= 0x1f) *ch += 0x84;
+			}
```
marzojr, line 3381: Maybe this should also be extracted to a separate function, just to be consistent with the above code blocks being extracted away.

marzojr (changes requested):

mapedit/studio.cc
Dragon-Baroque: I can extract them into separate functions.

However the second occurrence of the first conversion comes from the conversion failure recovery path.
Could you comment on whether that path is still needed ?

On the other hand I could place the whole failure recovery path as an else branch of the conversion success path, and place the byte move at the end of the complete if. Would be there once only then ...

marzojr (approved):

PR #339, opened 2023-04-08
Exult Studio uses the ICU to convert strings from and to UTF-8

Dragon-Baroque:
Commit 1 of 1
    .github/workflows/ci-linux.yml + ci-macos/yml + ci-windows.yml
      Add the ICU to the required components
    Makefile.mingw + configure.ac + mapedit/Makefile.am
      Add the ICU CFLAGS and LIBS to the build of Studio
    mapedit/exult_studio.glade
      Fix two windows that closed badly whex clicked on the cross
      Remove the not supported ISO-8859-16
    mapedit/npcedit.cc + shapeedit.cc + shapefile.cc + shapelst.cc
      Use the explicit conversion function, avoid string replicas
    mapedit/studio.h
      Replace the scaffolding classes utf8Str and codepageStr
        by explicit conversion functions
    mapedit/studio.cc [ the only source with ICU includes ]
      Implement the conversion functions, no longer any retry
      Use std::string as result

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -426,6 +426,14 @@ if test x$enable_android_apk = xno; then
 fi
 
 # ---------------------------------------------------------------------
+# Icu (for ES)
+# ---------------------------------------------------------------------
+
+if test x$enable_android_apk = xno; then
+    PKG_CHECK_MODULES(ICU, icu-uc, have_icu=yes, have_icu=no)
+fi
+
+# ---------------------------------------------------------------------
 # Gtk (for ES)
 # ---------------------------------------------------------------------
 
@@ -1049,8 +1057,13 @@ AC_MSG_CHECKING([whether to build Exult Studio])
 if test x$enable_exult_studio = xyes; then
 	AC_MSG_RESULT(yes)
 	if test x$have_gtk = xno; then
-		echo "Umm, but we don't have any Gtk+ stuff."
-		echo "Try again, either with Gtk-3.16 or newer, or with --disable-exult-studio"
+		echo "Exult Studio requires the GTK+ (aka the GIMP Tool Kit), but it is not installed."
+		echo "Please try again, either with the GTK+ installed as 3.16 or newer, or with '--disable-exult-studio'."
+		exit 1
+	fi
+	if test x$have_icu = xno; then
+		echo "Exult Studio requires the ICU (aka the International Components for Unicode), but it is not installed."
+		echo "Please try again, either with the ICU installed, or with '--disable-exult-studio'."
 		exit 1
 	fi
 	AM_CONDITIONAL(BUILD_STUDIO, true)
@@ -1285,6 +1298,9 @@ if test x$enable_android_apk = xno; then
 	if test x$enable_fluidsynth = xyes; then
 		PKG_CHECK_EXISTS(fluidsynth,
 			echo Fluidsynth................. : `$PKG_CONFIG --modversion fluidsynth`)
+	fi
+	if test x$have_icu = xyes; then
+		echo ICU........................ : `$PKG_CONFIG --modversion icu-uc`
 	fi
 	if test x$have_gtk = xyes; then
 		echo GLIB....................... : `$PKG_CONFIG --modversion glib-2.0`
--- a/mapedit/npcedit.cc
+++ b/mapedit/npcedit.cc
@@ -437,8 +437,8 @@ int ExultStudio::init_npc_window(
 	// Store address with window.
 	g_object_set_data(G_OBJECT(npcwin), "user_data", addr);
 	// Store name, ident, num.
-	const utf8Str utf8name(name.c_str());
-	set_entry("npc_name_entry", utf8name);
+	const std::string utf8name(convertToUTF8(name.c_str()));
+	set_entry("npc_name_entry", utf8name.c_str());
 	// (Not allowed to change npc#.).
 	set_entry("npc_num_entry", npc_num, true, false);
 	set_entry("npc_ident_entry", ident);
@@ -546,8 +546,7 @@ int ExultStudio::save_npc_window(
 	const int tx = -1;
 	const int ty = -1;
 	const int tz = -1;  // +++++For now.
-	const codepageStr locname(get_text_entry("npc_name_entry"));
-	const std::string name(locname);
+	const std::string name(convertFromUTF8(get_text_entry("npc_name_entry")));
 	const short npc_num = get_num_entry("npc_num_entry");
 	const short ident = get_num_entry("npc_ident_entry");
 	const int shape = get_num_entry("npc_shape");
--- a/mapedit/studio.cc
+++ b/mapedit/studio.cc
@@ -71,6 +71,9 @@ Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.
 #include <cstdlib>
 #include <memory>
 
+#include <unicode/ucnv.h>
+#include <unicode/ucnv_cb.h>
+
 using std::cerr;
 using std::cout;
 using std::endl;
@@ -1214,8 +1217,7 @@ C_EXPORT void on_gameselect_ok_clicked(
 		GtkTextIter endpos;
 		gtk_text_buffer_get_bounds(buff, &startpos, &endpos);
 		gchar *modmenu = gtk_text_iter_get_text(&startpos, &endpos);
-		const codepageStr menu(modmenu, "ASCII");
-		const string modmenustr = menu.get_str();
+		const string  modmenustr(convertFromUTF8(modmenu, "ASCII"));
 		g_free(modmenu);
 		if (modmenustr.empty())
 			return;
@@ -1313,10 +1315,10 @@ C_EXPORT void on_gameselect_gamelist_cursor_changed(
 
 		// Titles need to be displayable in Exult menu, hence should not
 		// have any extra characters.
-		const utf8Str title(modname.c_str(), "ASCII");
+		const string title(convertToUTF8(modname.c_str(), "ASCII"));
 		gtk_tree_store_append(model, &iter, nullptr);
 		gtk_tree_store_set(model, &iter,
-		                   0, title.get_str(),
+		                   0, title.c_str(),
 		                   1, j,
 		                   -1);
 	}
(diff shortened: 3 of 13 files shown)


configure.ac
```diff
@@ -1053,6 +1061,11 @@ if test x$enable_exult_studio = xyes; then
 		echo "Try again, either with Gtk-3.16 or newer, or with --disable-exult-studio"
 		exit 1
 	fi
+	if test x$have_icu = xno; then
+		echo "Umm, but we don't have any Icu stuff."
+		echo "Try again, either with Icu, or with --disable-exult-studio"
```
marzojr, line 1066: Maybe these two should say libicu instead of Icu? Or maybe ICU, all caps, like they do in their website?

mapedit/studio.cc
```diff
@@ -3164,228 +3160,302 @@ void ExultStudio::set_game_information(
...
-				force[pos] = '?';
-
-			CONV_ERROR(enc, "UTF-8");
-			_convstr = g_convert(force.c_str(), -1, "UTF-8", enc,
-			                     &bytes_read, &bytes_written, &error);
+	string cvt_str("");
+	if (is_7bit) {
+		cvt_str = src_str;
+		for (unsigned i = 0; i < strlen(cvt_str.c_str()); i++) {
+			cvt_str[ i ] = static_cast<char>(ToCP437[ static_cast<unsigned char>(cvt_str[ i ]) ]);
```
marzojr, line 3434: Maybe this would be better:
```c++
for (char& c : cvt_str) {
	c = static_cast<char>(ToCP437[static_cast<unsigned char>(c)]);
}
```

mapedit/studio.cc
```diff
@@ -3164,228 +3160,302 @@ void ExultStudio::set_game_information(
...
-			const char illegal = force[bytes_read];
-			size_t pos;
-
-			while ((pos = force.find(illegal)) != string::npos)
-				force[pos] = '?';
-
-			CONV_ERROR(enc, "UTF-8");
-			_convstr = g_convert(force.c_str(), -1, "UTF-8", enc,
-			                     &bytes_read, &bytes_written, &error);
+	string cvt_str("");
```
marzojr, line 3430: Remove `("")`.

mapedit/studio.cc
```diff
@@ -3164,228 +3160,302 @@ void ExultStudio::set_game_information(
...
+	}
 
-		if (!_convstr) {
-			CONV_ERROR(enc, "UTF-8");
-			// Conversion still failed; try lossy conversion.
-			const char fallback = '?';
-			_convstr = g_convert_with_fallback(force.c_str(), -1, "UTF-8", enc,
-			                                   &fallback, &bytes_read, &bytes_written, &error);
-		}
+	string tgt_str(1 + strlen(src_str)*3, ' ');
```
marzojr, line 3439: Should probably be `*4` to be safe. You never know what users might be typing, and getting random crashes because someone types an emoji would not be nice.

marzojr: Also, instead of calling `strlen(src_str)` numerous times, you should call it and assign the result to a const variable.

configure.ac
Dragon-Baroque: I just copied the check from GTK+ just above, and it uses `Gtk+` not `GTK+`. Should we change both to all caps ?

mapedit/studio.cc
```diff
@@ -3164,228 +3160,302 @@ void ExultStudio::set_game_information(
...
+
+	auto src_ptr = src_str;
+	auto src_end = src_ptr + strlen(src_str) + 1;
+	auto tgt_ptr = const_cast<char *>(tgt_str.c_str());
```
marzojr, line 3443: It is more appropriate to use `tgt_str.data()` in this case; after c++17, the const_cast would not even be needed anymore with `tgt_str.data()`.

mapedit/studio.cc
```diff
@@ -3164,228 +3160,302 @@ void ExultStudio::set_game_information(
...
-			else if (*ch <= 0x9f) ;
-			// a0 -> 19 á:a acute.
-			else if (*ch <= 0xa0) *ch -= 0x87;
-			// a1 -> 1d í:i acute,      a2 -> 1e ó:o acute,      a3 -> 1f ú:u acute.
-			else if (*ch <= 0xa3) *ch -= 0x84;
-			else if (*ch <= 0xe0) ;
-			// e1 -> 1c ß:s sharp.
-			else if (*ch <= 0xe1) *ch -= 0xc5;
+		for (unsigned i = 0; i < strlen(tgt_str.c_str()); i++) {
+			tgt_str[ i ] = static_cast<char>(FromCP437[ static_cast<unsigned char>(tgt_str[ i ]) ]);
```
marzojr, line 3369: Should probably be
```c++
for (char& c : tgt_str) {
	c = static_cast<char>(FromCP437[static_cast<unsigned char>(c)]);
}
```

mapedit/studio.cc
```diff
@@ -3164,228 +3160,302 @@ void ExultStudio::set_game_information(
...
+	if (setupEncConverter( enc )  == false) {
+		return "";
+	}
+
+	string tgt_str(1 + strlen(src_str)*1, ' ');
+
+	auto src_ptr = src_str;
+	auto src_end = src_ptr + strlen(src_str) + 1;
+	auto tgt_ptr = const_cast<char *>(tgt_str.c_str());
+	auto tgt_end = tgt_str.c_str() + strlen(src_str)*1 + 1;
```
marzojr, line 3324: `tgt_str.c_str() + strlen(src_str)*3 + 1;` -> `tgt_ptr + tgt_str.size() + 1;`

mapedit/studio.cc
```diff
@@ -3164,228 +3160,302 @@ void ExultStudio::set_game_information(
...
+	}
+	if (setupEncConverter( enc )  == false) {
+		return "";
+	}
+
+	string tgt_str(1 + strlen(src_str)*1, ' ');
+
+	auto src_ptr = src_str;
+	auto src_end = src_ptr + strlen(src_str) + 1;
+	auto tgt_ptr = const_cast<char *>(tgt_str.c_str());
```
marzojr, line 3323: It is more appropriate to use `tgt_str.data()` in this case; after c++17, the const_cast would not even be needed anymore with `tgt_str.data()`.

mapedit/studio.cc
```diff
@@ -3164,228 +3160,302 @@ void ExultStudio::set_game_information(
...
-			const char fallback = '?';
-			_convstr = g_convert_with_fallback(force.c_str(), -1, "UTF-8", enc,
-			                                   &fallback, &bytes_read, &bytes_written, &error);
-		}
+	string tgt_str(1 + strlen(src_str)*3, ' ');
+
+	auto src_ptr = src_str;
+	auto src_end = src_ptr + strlen(src_str) + 1;
+	auto tgt_ptr = const_cast<char *>(tgt_str.c_str());
+	auto tgt_end = tgt_str.c_str() + strlen(src_str)*3 + 1;
```
marzojr, line 3444: `tgt_str.c_str() + strlen(src_str)*3 + 1;` -> `tgt_ptr + tgt_str.size() + 1;`

mapedit/studio.cc
```diff
@@ -3164,228 +3160,302 @@ void ExultStudio::set_game_information(
...
-		}
-	}
+	if (setupUtfConverter("UTF8") == false) {
+		return "";
+	}
+	if (setupEncConverter( enc )  == false) {
+		return "";
+	}
+
+	string tgt_str(1 + strlen(src_str)*1, ' ');
```
marzojr, line 3319: Instead of calling `strlen(src_str)` numerous times, you should call it and assign the result to a const variable.

marzojr, line 3319: This also should probably just use '\0' instead of ' '.

mapedit/studio.cc
marzojr, line 3439: This also should probably just use '\0' instead of ' '.

mapedit/npcedit.cc
```diff
@@ -437,8 +437,8 @@ int ExultStudio::init_npc_window(
 	// Store address with window.
 	g_object_set_data(G_OBJECT(npcwin), "user_data", addr);
 	// Store name, ident, num.
-	const utf8Str utf8name(name.c_str());
-	set_entry("npc_name_entry", utf8name);
+	std::string utf8name(convertToUTF8(name.c_str()));
```
marzojr, line 440: This applies in general, but picking this specific case: why not `const string`?

marzojr (changes requested):

configure.ac
marzojr: Might as well.

marzojr (approved):

mapedit/studio.cc
```diff
@@ -3164,228 +3160,304 @@ void ExultStudio::set_game_information(
...
+	}
+	if (setupEncConverter( enc )  == false) {
+		return "";
+	}
+
+	string tgt_str(1 + src_len*1, '\0');
+
+	auto src_ptr = src_str;
+	auto src_end = src_ptr + src_len + 1;
+	auto tgt_ptr = tgt_str.data();
```
marzojr, line 3324: I should have been clearer: with c++17 it works without the const_cast, but until then, we need it.

mapedit/studio.cc
```diff
@@ -3164,228 +3160,304 @@ void ExultStudio::set_game_information(
...
-			// Conversion still failed; try lossy conversion.
-			const char fallback = '?';
-			_convstr = g_convert_with_fallback(force.c_str(), -1, "UTF-8", enc,
-			                                   &fallback, &bytes_read, &bytes_written, &error);
-		}
+	string tgt_str(1 + src_len*4, '\0');
+
+	auto src_ptr = src_str;
+	auto src_end = src_ptr + src_len + 1;
+	auto tgt_ptr = tgt_str.data();
```
marzojr, line 3445: I should have been clearer: with c++17 it works without the const_cast, but until then, we need it.

marzojr (changes requested):

mapedit/studio.cc
Dragon-Baroque: Yes, I understood that statement.
But the macOS and Linux builds _are_ C++17 while - and I have realized that just a moment ago - the Windows build remains as C++ 14.
Could you explain that discrepancy ?

marzojr: We haven't "officially" switched to c++17 yet. I suppose we should sometime soon.

marzojr (approved):

PR #366, opened 2023-07-10
Add support for Haiku's layout

Begasus:
Save configuration file and savehome_dir to ~/config/settings/exult 
Use data from ~/config/non-packaged/data/exult (system data folder is read-only so users can't install their game data there)


Begasus:
Thanks!

DominusExult:
My pleasure ;)

Issue #416, opened 2023-12-26
clang 17 and 16 warnings


nunotexbsd:
I've noticed a lot of warnings on clang 17 and 16 that might upstream be interested in take a look.
Tested with 1.9.0.20231225 on FreeBSD.

```
2814 warnings generated.
mv -f .deps/paledit.Tpo .deps/paledit.Po
2802 warnings generated.
mv -f .deps/ucbrowse.Tpo .deps/ucbrowse.Po
2806 warnings generated.
mv -f .deps/shapegroup.Tpo .deps/shapegroup.Po
2810 warnings generated.
mv -f .deps/studio.Tpo .deps/studio.Po
2802 warnings generated.
mv -f .deps/shapefile.Tpo .deps/shapefile.Po
2810 warnings generated.
mv -f .deps/shapeedit.Tpo .deps/shapeedit.Po
2826 warnings generated.
```

clang17 [log](https://people.freebsd.org/~eduardo/logs/clang_warnings/exult-devel/15amd64_exult-devel-1.9.0.20231225%2C1.log) ~74MB
clang16 [log](https://people.freebsd.org/~eduardo/logs/clang_warnings/exult-devel/140amd64_exult-devel-1.9.0.20231225%2C1.log) ~71MB
clang14 [log](https://people.freebsd.org/~eduardo/logs/clang_warnings/exult-devel/132amd64_exult-devel-1.9.0.20231225%2C1.log) ~ 631KB

Dragon-Baroque:
This is shortly mentioned in issue #394.

All GTK callbacks, whatever their true signature, are cast as `GCallback` or `(void)` ( without arguments ).
All these casts fail with the new warning `-Wcast-function-type-strict` from CLang 16, that is dragged by `-Wcast-function-type`.
Until we reach a decision, I suggest you alter `./configure.ac` by adding `-Wno-cast-function-type-strict` after `-Wcast-function-type`. 

Now if there remain other differences in CLang between 14, 16, and 17 that the `-Wcast-function-type-strict` warnings, we are interested.

nunotexbsd:
@Dragon-Baroque 

I've follow your sugestion and logs back to normal. 

- affected clang 17,16 (with -Wno-cast-function-type-strict added)
- unaffected clang 15,14,13

exult 1.8 and 20231225 [logs](https://people.freebsd.org/~eduardo/logs/clang_warnings/exult-devel/Wno-cast-function-type-strict/)

Thanks!

[...]

Issue #394, opened 2023-09-03
macOS compiler warnings


DominusExult:
Some new compiler warnings on macOS:

```
zip.cc:426:9: warning: use of old-style cast [-Wold-style-cast]
                err = deflateInit2(&file->ci.stream, level,
                      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/opt/x86_64/include/zlib.h:1816:51: note: expanded from macro 'deflateInit2'
                        (strategy), ZLIB_VERSION, (int)sizeof(z_stream))
                                                  ^    ~~~~~~~~~~~~~~~~
```

```
unzip.cc:927:9: warning: use of old-style cast [-Wold-style-cast]
                err = inflateInit2(&pfile_in_zip_read_info->stream, -MAX_WBITS);
                      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/opt/x86_64/include/zlib.h:1819:25: note: expanded from macro 'inflateInit2'
                        (int)sizeof(z_stream))
                        ^    ~~~~~~~~~~~~~~~~
```

```
Table.cc:44:15: warning: variable 'i' set but not used [-Wunused-but-set-variable]
        unsigned int i = 0;
                     ^
```
```
In file included from U7file.cc:17:
./U7file.h:188:2: warning: definition of implicit copy assignment operator for 'File_data' is deprecated because it has a user-declared copy constructor [-Wdeprecated-copy]
        File_data(const File_data &other) noexcept = default;
```
```
U7file.cc:34:8: note: in instantiation of function template specialization 'std::vector<File_data>::emplace_back<const File_spec &>' requested here
        files.emplace_back(spec);
```
```
Gamemenu_gump.cc:55:20: warning: unused variable 'quitmenutext' [-Wunused-variable]
static const char *quitmenutext = "Quit to Menu";
```

Dragon-Baroque:
I have pushed and merged PR #413 to display the version of the compiler in the logs, to help understand how to silent its warnings.

@marzojr, at this point, could we consider that Exult will never support any other compiler than GCC and CLang, and reduce the compiler list `cxx_compilers` in `configure.ac` accordingly ? I mean `cxx_compilers="g++ clang++"` except for macOS / iOS `cxx_compilers="clang++ g++"`.

@DominusExult, I cannot silent the warnings of CLang without increasing the warnings of GCC.
[...]

PR #419, opened 2023-12-30
Silence more compiler warnings

Dragon-Baroque:
- audio/midi_drivers/timidity_sdl_c.cpp + timidity.h + .cpp : Replace the last (v)sprintf by a (v)snprintf,
- configure.ac : Rework the failing FreeType2 installation check, made Gen_shadow useless, Enable -Wdeprecated,
- gamemgr/bggame.cc : Remove a declaration duplicated [ spotted by -Wdeprecated ],
- pathfinder/Pathfinder.h : Fix a new C++ deprecated warning [ -Wdeprecated ]. @marzojr I am not sure of this fix, would you please review it ?
- many with #include <SDL.h> : Silent -Wzero-as-null-pointer-constant,
- gamewin.cc : Replace some hardcoded string size by sizeof, Remove useless store of zero at end of sprintf because snprintf always provides a ( possibly truncated ) zero terminated string,
- gumps/VideoOptions_gump.cc : Enlarge a sprintf target string to ensure it has space enough.
- audio/OggAudioSample.h : Silence Vorbis warnings
- audio/midi_drivers/MT32EmuMidiDriver.h : Silence MT32Emu warnings
- mapedit/Studio.cc : Silence ICU warnings

For issues #394 and #416

### DIFF
--- a/gumps/VideoOptions_gump.cc
+++ b/gumps/VideoOptions_gump.cc
@@ -29,6 +29,7 @@
 #ifdef __GNUC__
 #	pragma GCC diagnostic push
 #	pragma GCC diagnostic ignored "-Wold-style-cast"
+#	pragma GCC diagnostic ignored "-Wzero-as-null-pointer-constant"
 #endif    // __GNUC__
 #include <SDL.h>
 #ifdef __GNUC__
@@ -206,9 +207,7 @@ void VideoOptions_gump::rebuild_dynamic_buttons() {
 		// the text arrays is freed by the destructor of the button
 		std::vector<std::string> scalingtext;
 		for (int i = 0; i < num_scales; i++) {
-			char buf[10];
-			snprintf(buf, sizeof(buf), "x%d", i + 1);
-			scalingtext.emplace_back(buf);
+			scalingtext.push_back('x' + std::to_string(i + 1));
 		}
 		buttons[id_scaling] = std::make_unique<VideoTextToggle>(this, &VideoOptions_gump::toggle_scaling,
 		        std::move(scalingtext), scaling, colx[2], rowy[4], 74);
--- a/audio/Audio.h
+++ b/audio/Audio.h
@@ -26,6 +26,7 @@
 #ifdef __GNUC__
 #	pragma GCC diagnostic push
 #	pragma GCC diagnostic ignored "-Wold-style-cast"
+#	pragma GCC diagnostic ignored "-Wzero-as-null-pointer-constant"
 #endif    // __GNUC__
 #include <SDL.h>
 #ifdef __GNUC__
--- a/audio/AudioMixer.cc
+++ b/audio/AudioMixer.cc
@@ -27,6 +27,7 @@ Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.
 #ifdef __GNUC__
 #	pragma GCC diagnostic push
 #	pragma GCC diagnostic ignored "-Wold-style-cast"
+#	pragma GCC diagnostic ignored "-Wzero-as-null-pointer-constant"
 #endif    // __GNUC__
 #include <SDL.h>
 #ifdef __GNUC__
--- a/audio/OggAudioSample.cc
+++ b/audio/OggAudioSample.cc
@@ -24,6 +24,7 @@ Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.
 #ifdef __GNUC__
 #	pragma GCC diagnostic push
 #	pragma GCC diagnostic ignored "-Wold-style-cast"
+#	pragma GCC diagnostic ignored "-Wzero-as-null-pointer-constant"
 #endif    // __GNUC__
 #include <SDL.h>
 #ifdef __GNUC__
--- a/audio/OggAudioSample.h
+++ b/audio/OggAudioSample.h
@@ -19,7 +19,15 @@ Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.
 #define OggAudioSample_H
 
 #include "AudioSample.h"
+#ifdef __GNUC__
+#	pragma GCC diagnostic push
+#	pragma GCC diagnostic ignored "-Wold-style-cast"
+#	pragma GCC diagnostic ignored "-Wzero-as-null-pointer-constant"
+#endif    // __GNUC__
 #include <vorbis/vorbisfile.h>
+#ifdef __GNUC__
+#	pragma GCC diagnostic pop
+#endif    // __GNUC__
 #include <memory>
 
 namespace Pentagram {
--- a/audio/WavAudioSample.cc
+++ b/audio/WavAudioSample.cc
@@ -24,6 +24,7 @@ Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.
 #ifdef __GNUC__
 #	pragma GCC diagnostic push
 #	pragma GCC diagnostic ignored "-Wold-style-cast"
+#	pragma GCC diagnostic ignored "-Wzero-as-null-pointer-constant"
 #endif    // __GNUC__
 #include <SDL.h>
 #ifdef __GNUC__
--- a/audio/midi_drivers/LowLevelMidiDriver.h
+++ b/audio/midi_drivers/LowLevelMidiDriver.h
@@ -33,6 +33,7 @@ class XMidiSequence;
 #ifdef __GNUC__
 #	pragma GCC diagnostic push
 #	pragma GCC diagnostic ignored "-Wold-style-cast"
+#	pragma GCC diagnostic ignored "-Wzero-as-null-pointer-constant"
 #endif    // __GNUC__
 #include <SDL.h>
 #ifdef __GNUC__
--- a/audio/midi_drivers/MT32EmuMidiDriver.h
+++ b/audio/midi_drivers/MT32EmuMidiDriver.h
@@ -24,11 +24,22 @@ Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.
 
 #include "LowLevelMidiDriver.h"
 
+#ifdef __GNUC__
+#	pragma GCC diagnostic push
+#	pragma GCC diagnostic ignored "-Wundef"
+#	pragma GCC diagnostic ignored "-Wzero-as-null-pointer-constant"
+#	if defined(__llvm__) || defined(__clang__)
+#		pragma GCC diagnostic ignored "-Wmacro-redefined"
+#	endif
+#endif    // __GNUC__
 #ifdef __IPHONEOS__
 	#include <mt32emu.h>
 #else
 	#include <mt32emu/mt32emu.h>
 #endif
+#ifdef __GNUC__
+#	pragma GCC diagnostic pop
+#endif    // __GNUC__
 
 namespace MT32Emu {
 	class Synth;
--- a/audio/midi_drivers/timidity/timidity.cpp
+++ b/audio/midi_drivers/timidity/timidity.cpp
@@ -427,7 +427,7 @@ void Timidity_DeInit()
 }
 
 
-char timidity_error[1024] = "";
+char timidity_error[TIMIDITY_ERROR_SIZE] = "";
 char *Timidity_Error()
 {
 	return timidity_error;
--- a/audio/midi_drivers/timidity/timidity.h
+++ b/audio/midi_drivers/timidity/timidity.h
@@ -32,6 +32,8 @@ namespace NS_TIMIDITY {
 
 struct MidiSong;
 
+#define TIMIDITY_ERROR_SIZE	1024
+
 //extern int Timidity_Init(int rate, int format, int channels, int samples);
 extern char *Timidity_Error();
 extern void Timidity_SetVolume(int volume);
--- a/audio/midi_drivers/timidity/timidity_sdl_c.cpp
+++ b/audio/midi_drivers/timidity/timidity_sdl_c.cpp
@@ -108,7 +108,7 @@ static int cmsg(int type, int verbosity_level, const char *fmt, ...)
 	    ctl.verbosity<verbosity_level)
 		return 0;
 	va_start(ap, fmt);
-	vsprintf(timidity_error, fmt, ap);
+	vsnprintf(timidity_error, TIMIDITY_ERROR_SIZE, fmt, ap);
 	va_end(ap);
 	//perr.printf ("%s\n", timidity_error);
 	return 0;
--- a/audio/soundtest.cc
+++ b/audio/soundtest.cc
@@ -23,6 +23,7 @@
 #ifdef __GNUC__
 #	pragma GCC diagnostic push
 #	pragma GCC diagnostic ignored "-Wold-style-cast"
+#	pragma GCC diagnostic ignored "-Wzero-as-null-pointer-constant"
 #endif    // __GNUC__
 #include <SDL.h>
 #ifdef __GNUC__
--- a/browser.cc
+++ b/browser.cc
@@ -23,6 +23,7 @@
 #ifdef __GNUC__
 #	pragma GCC diagnostic push
 #	pragma GCC diagnostic ignored "-Wold-style-cast"
+#	pragma GCC diagnostic ignored "-Wzero-as-null-pointer-constant"
 #endif    // __GNUC__
 #include <SDL.h>
 #ifdef __GNUC__
--- a/cheat.cc
+++ b/cheat.cc
@@ -32,6 +32,7 @@ using std::setfill;
 #ifdef __GNUC__
 #	pragma GCC diagnostic push
 #	pragma GCC diagnostic ignored "-Wold-style-cast"
+#	pragma GCC diagnostic ignored "-Wzero-as-null-pointer-constant"
 #endif    // __GNUC__
 #include <SDL.h>
 #ifdef __GNUC__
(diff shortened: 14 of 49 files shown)


Dragon-Baroque:
@marzojr, this should be your expertise :

`gamemgr/bggame.cc` duplicate declaration reinstated because ci-windows failed without it. With CLang++ 17 this causes anew the warning :
```
bggame.cc:1143:62: warning: out-of-line definition of constexpr static data member is redundant in C++17 and is deprecated [-Wdeprecated-redundant-constexpr-static-def]
 1143 | constexpr const Hand_Handler::HandlerScriptOps Hand_Handler::HandlerScript[];
      |                                                              ^
```
`pathfinder/Pathfinder.h` : At first, I added the copy constructor `Pathfinder_client(const Pathfinder_client&) = default;` to silent the CLang C++ 17 warning 
```
./pathfinder/PathFinder.h:32:10: warning: definition of implicit copy constructor for 'Pathfinder_client' is deprecated because it has a user-declared destructor [-Wdeprecated-copy-with-dtor]
   32 |         virtual ~Pathfinder_client() = default;
      |                 ^
```
but CodeQL requires that I add also the copy assignment operator `Pathfinder_client& operator=(const Pathfinder_client&) = default;`. In a similar situation, for class `File_data`, you merely removed the copy constructor in commit 229419a1.

pathfinder/PathFinder.h
```diff
@@ -30,6 +30,8 @@ class Pathfinder_client {
 public:
 	Pathfinder_client(int mf) : move_flags(mf) {  }
 	virtual ~Pathfinder_client() = default;
+	Pathfinder_client(const Pathfinder_client&) = default;
+	Pathfinder_client& operator=(const Pathfinder_client&) = default;
```
marzojr, line 34: I recommend changing these two to `= delete;` and adding two more for the move constructor/assignment:

```c++
	Pathfinder_client(const Pathfinder_client&) = delete;
	Pathfinder_client(const Pathfinder_client&) = delete;
	Pathfinder_client& operator=(Pathfinder_client&&) = delete;
	Pathfinder_client& operator=(Pathfinder_client&&) = delete;
```
The idea being that copying or moving a class with virtual members is almost always wrong and will lead to slicing.

gumps/VideoOptions_gump.cc
```diff
@@ -206,7 +207,7 @@ void VideoOptions_gump::rebuild_dynamic_buttons() {
 		// the text arrays is freed by the destructor of the button
 		std::vector<std::string> scalingtext;
 		for (int i = 0; i < num_scales; i++) {
-			char buf[10];
+			char buf[16];
 			snprintf(buf, sizeof(buf), "x%d", i + 1);
 			scalingtext.emplace_back(buf);
```
marzojr, line 212: I would recommend to do away with `buf` altogether and replace the entire loop body by this:
```c++
scalingtext.push_back('x' + std::to_string(i + 1));
```

marzojr (changes requested):

pathfinder/PathFinder.h
Dragon-Baroque: It does not work, because the copy constructor is used by `schedule.cc` on the derived class `Actor_pathfinder`.
```
schedule.cc:4474:27: error: call to implicitly-deleted copy constructor of 'Actor_pathfinder_client'
 4474 |                 Actor_pathfinder_client COST = (GAME_SI ? cost : cost2);
      |                                         ^      ~~~~~~~~~~~~~~~~~~~~~~~~
./paths.h:36:33: note: copy constructor of 'Actor_pathfinder_client' is implicitly deleted because base class 'Pathfinder_client' has a deleted copy constructor
   36 | class Actor_pathfinder_client : public Pathfinder_client {
      |                                 ^
./pathfinder/PathFinder.h:33:2: note: 'Pathfinder_client' has been explicitly marked deleted here
   33 |         Pathfinder_client(const Pathfinder_client&) = delete;
      |         ^
1 error generated.
```
I made the change :
```
-       Pathfinder_client(const Pathfinder_client&) = default;
-       Pathfinder_client& operator=(const Pathfinder_client&) = default;
+       Pathfinder_client(const Pathfinder_client&) = delete;
+       Pathfinder_client(Pathfinder_client&&) = delete;
+       Pathfinder_client& operator=(const Pathfinder_client&) = delete;
+       Pathfinder_client& operator=(Pathfinder_client&&) = delete;
```

marzojr:
> bggame.cc:1143:62: warning: out-of-line definition of constexpr static data member is redundant in C++17 and is deprecated [-Wdeprecated-redundant-constexpr-static-def]
>  1143 | constexpr const Hand_Handler::HandlerScriptOps Hand_Handler::HandlerScript[];
>       |                                                              ^
> ```

This one is annoying; in c++14 we don't have access to inline variables or we could add `inline` somewhere in the declaration in line 1075 (inline variables is a c++17feature); but c++17 also relaxed `constexpr` restrictions and made `static` `constexpr` member variables implicitly `inline`. So the only way to really handle this is ugly:

```c++
#if __cplusplus < 201703L
constexpr const Hand_Handler::HandlerScriptOps Hand_Handler::HandlerScript[];
#endif
```

> ./pathfinder/PathFinder.h:32:10: warning: definition of implicit copy constructor for 'Pathfinder_client' is deprecated because it has a user-declared destructor [-Wdeprecated-copy-with-dtor]
>    32 |         virtual ~Pathfinder_client() = default;
>       |                 ^
> ```
> 
> but CodeQL requires that I add also the copy assignment operator `Pathfinder_client& operator=(const Pathfinder_client&) = default;`. In a similar situation, for class `File_data`, you merely removed the copy constructor in commit [229419a](https://github.com/exult/exult/commit/229419a119a9d0b0ef652329c0e7cb0fbdd36d3e).

There are a few things to consider about this one; I briefly mentioned in the comment, but it is good to have a more complete answer here.

Under normal circumstances, if you do not need to, you should not declare/default/delete any of destructor, copy constructor, move constructor, copy assignment, and move assignment. This is called "the rule of zero". You will then have access to compiler-generated versions of all of these.

Declaring some of those disables some of the others; the rules for this are a bit complex, but the general rule is: if you declare/delete any of them, you are probably doing something in this function which is not what the compiler would generate by itself; therefore, you should declare/delete all the others. This is called "the rule of five" (it was "the rule of 3" in c++98/c++03).

Classes or structures with `virtual` members are an important exception; it is almost always wrong to have such a class without a `virtual` destructor. But it is also almost always wrong to define a copy/move constructor/assignment for such classes: when these would be called, you would generally be using a pointer at some point of the hierarchy, and the version called would correspond to the static type of the pointer. This would lead to slicing, where only part of the object is copied or moved. So it is almost always better to `= delete;` these.

`File_data` was a non-virtual class, and it did just copy of the data members. So deleting the definition of the copy constructor was enough to implicitly generate all other special member functions.

`Pathfinder_client`, on the other hand, is a virtual base class. Therefore the right way to handle it is to delete all of its special member functions.

marzojr:
> This one is annoying; in c++14 we don't have access to inline variables or we could add `inline` somewhere in the declaration in line 1075 (inline variables is a c++17feature); but c++17 also relaxed `constexpr` restrictions and made `static` `constexpr` member variables implicitly `inline`. So the only way to really handle this is ugly:
> 
> ```c++
> #if __cplusplus < 201703L
> constexpr const Hand_Handler::HandlerScriptOps Hand_Handler::HandlerScript[];
> #endif
> ```
I was thinking. If the HandlerScriptOps enumand the HandlerScriptOps variable are both moved outside of the class, we can also get rid of the warning. This is, in my opinion, a better way to handle the issue.

Dragon-Baroque:
Fixed `gumps/VideoOptions_gump.cc` as you suggested.

Moved `HandlerScriptsOps` and `HandlerScript` out of the class `Hand_Handler` as you suggested. Compiles without warnings.

Reverted `pathfinder/Pathfinder.h` to its state before the PR. I am trying to understand the logic of those virtual destructors.

Dragon-Baroque:
Please tell me if I am correct :

- You need to code a user destructor if your class contains fields like `char *` whose target should be freed, and ( rule of five ) you need to code the copy and move constructors and assignment operators [ that you actually use ].
- You need to declare a default destructor if your class contains fields like `std::string` or any instance of a class that has its own destructor method, or any pointer to instance of a class whether or not it has a destructor, and ( rule of five ) you need to declare default copy and move constructors and assignment operators [ that you actually use ].

- You need to declare your methods `virtual` - and that includes the destructor if it is needed - if the instances of the class and its derived classes use polymorphism ( the instances of a derived class are stored and used as instances of the base class ).

Now when, like some time ago, you do not declare the default constructor, it appears as if C++ generates _some_ generic constructor and likewise some generic copy and move constructors and assignment operators. Are they made virtual ? Do their code differ from their `default` equivalents ? Because by now, a mismatch between a default declared destructor and an implicit ( ie non declared ) copy constructor, say, only raises a warning and then only if the proper warning is enabled.

Indeed, you could silence the warning about `File_data` because you could avoid declaring the virtual default destructor - the class had not inheritance and thus used no polymorphism. But yet, because of the `File_spec` and `std::unique_ptr<IDataSource>` fields, the class needs a generic destructor and a generic copy constructor. Could you remove the default constructor because it does not need to be virtual ?

wench:
> > This one is annoying; in c++14 we don't have access to inline variables or we could add `inline` somewhere in the declaration in line 1075 (inline variables is a c++17feature); but c++17 also relaxed `constexpr` restrictions and made `static` `constexpr` member variables implicitly `inline`. So the only way to really handle this is ugly:
> > ```c++
> > #if __cplusplus < 201703L
> > constexpr const Hand_Handler::HandlerScriptOps Hand_Handler::HandlerScript[];
> > #endif
> > ```
> 
> I was thinking. If the HandlerScriptOps enumand the HandlerScript variable are both moved outside of the class, we can also get rid of the warning. This is, in my opinion, a better way to handle the issue.

Any particular reason we cannot up the C++ version to 17? Surely every relevant platform has a compiler that supports it by now as the standard is over 6 years old

marzojr:
@Dragon-Baroque:

> Reverted `pathfinder/Pathfinder.h` to its state before the PR. I am trying to understand the logic of those virtual destructors.

Let me take a look at this one and I will report back.

>     * You need to code a user destructor if your class contains fields like `char *` whose target should be freed, and ( rule of five ) you need to code the copy and move constructors and assignment operators [ that you actually use ].

Correct. The logic behind the rule of 5 being that doing a copy using the default compiler-generated copy constructor/assignment (basically a copy of each member, and sometimes a memcpy of the whole class) would end up with two copies of the class with the same `char *`, and both copies of the class would be deleting the `char *` in the destructor, a double-delete. Similarly for move constructor/assignment: the compiler-generated version is basically a move from each member (and sometimes a memcpy of the whole class).

>     * You need to declare a default destructor if your class contains fields like `std::string` or any instance of a class that has its own destructor method, or any pointer to instance of a class whether or not it has a destructor, and ( rule of five ) you need to declare default copy and move constructors and assignment operators [ that you actually use ].

Yes and no: "no" for the first part, as you do not need to declare a destructor at all if all that the destructor needs to do is go through all member variables and do and call their destructors (with destructors of fundamental types being no-ops). So the following is a class that will correctly destruct all its members, and have a compiler-generated destructor, copy constructor/assignment, and move constructor/assignment:

```c++
struct A {
    std::string name;
};
```

The following is a class that will correctly destruct all its members, and have a compiler-generated destructor, and move constructor/assignment, but with deleted copy constructor/assignment (because `std::unique_ptr` deletes these):

```c++
struct B;
struct C {
    std::unique_ptr<B> value;
};
```

But "yes" for the "middle" part, in that you should declare/default/delete all of destructor, copy constructor/assignment, and move constructor/assignment if you declare/default/delete any of them (and note that you can declare some, default some, and delete the others, but you need to think about why you are doing that -- for example, the `std::unique_ptr` is a unique owner of a resource, and so it makes sense to have non-default non-deleted move constructor/assignment (transfer of ownership), deleted copy constructor/assignment (ownership is unique), and non-default non-deleted destructor (for releasing the resource).

And "no" for the tail end: you should declare all of them *even if you do not use them*: this way, you will prevent accidental mistakes and will get an error during compilation, instead of potential memory corruption, or other issues due to partial object copies during runtime.

>     * You need to declare your methods `virtual` - and that includes the destructor if it is needed - if the instances of the class and its derived classes use polymorphism ( the instances of a derived class are stored and used as instances of the base class ).

If by "your methods" you mean "any of destructor, copy constructor/assignment, and move constructor/assignment", then no -- of these, only the destructor can be virtual (you get a compilation error if you try making any of the others virtual, and this is the issue). If you mean "destructor or any methods other than copy constructor/assignment, and move constructor/assignment", then yes -- and, generally, you just want to `= delete;` these four.

As an exception, if you have a virtual `clone` method in a class hierarchy, you could maybe use non-deleted, protected, copy constructor/assignment, and move constructor/assignment to be use these in `clone`. But again, it requires thinking about what you are doing to make sure you are doing it right.

> Now when, like some time ago, you do not declare the default constructor, it appears as if C++ generates _some_ generic constructor and likewise some generic copy and move constructors and assignment operators. Are they made virtual ?

Unless a base class has a `virtual` method with the same name and function signature (i.e., an override), C++ will never make anything on a derived class `virtual` unless you tell it to. Moreover, copy constructor/assignment, and move constructor/assignment can never be virtual.

(And as an aside, these implicitly `virtual` method overrides should be marked as `override` or `final` -- but *not* `virtual` -- so you can get compiler errors if they don't actually override anything from the base class -- for example, because you made a typo in the name, or got the function signature wrong, or are doing a refactor and forgot to change one class in the hierarchy.)

> Do their code differ from their `default` equivalents ?

The implicit and `= default;` versions are generally identical in codegen: the compiler will use a memcpy when possible (e.g., if on a non-polymorphic class on which all member variables have "trivial" copy constructor/assignment), or a memberwise copy otherwise.

> Because by now, a mismatch between a default declared destructor and an implicit ( ie non declared ) copy constructor, say, only raises a warning and then only if the proper warning is enabled.

The issue is that there are some cases in which the mismatch might be warranted, and there is a ton of code out there which may have such mismatches (sometimes intentionally). A prime example is COM on Windows -- the base class/interface of the COM hierarchy has virtual member functions, but not a virtual destructor or any data members: the idea being that COM has reference counting, and you must implement AddRef and Release virtual member functions which will use the correct (statically typed) destructor, instead of directly using `delete` on pointers to COM objects.

But again, it is something that requires thinking about to make sure you are doing correctly; and this is the point of the warning, and the reason why it is optional.

> Indeed, you could silence the warning about `File_data` because you could avoid declaring the virtual default destructor - the class had not inheritance and thus used no polymorphism. But yet, because of the `File_spec` and `std::unique_ptr<IDataSource>` fields, the class needs a generic destructor and a generic copy constructor. Could you remove the default constructor because it does not need to be virtual ?

I am assuming you are talking about `U7file`? It needs a virtual destructor because it is at the root of a polymorphic hierarchy. It could get away with not declaring a destructor at all except for this fact. And now that I think about it, its move constructor/assignment should be deleted too, instead of defaulted.

@wench 
> Any particular reason we cannot up the C++ version required by Exult to 17? Surely every relevant platform has a compiler that supports it by now as the standard is over 6 years old

I would agree even jumping straight to c++20, but compiler support for that is still spotty primarily due to Apple Clang, but also due to clang itself ([source](https://en.cppreference.com/w/cpp/compiler_support/20)).

But yes, you are right, we could (should?) just swap to c++17 -- we even have GCC 12.2 for DOS (via [DJGPP](https://www.delorie.com/pub/djgpp/current/v2gnu/)), so any platforms without full (or near full) support for c++17 is deader than DOS. We only have to take some care with library ([source](https://en.cppreference.com/w/cpp/compiler_support/17)):

- libc++ does not have full c++17 support yet;
- Apple Clang also does not have full c++17 support (and it depends on the Xcode version).

The Android NDK is currently using clang 9.0.9; but we need to check the versions of Xcode being used on the supported versions of Apple stuff. @DominusExult can you comment on this?

For what is worth, unless we are looking at a very ancient version of Xcode, I don't think that the missing support will be an issue for us.

marzojr (approved):

DominusExult:
` but we need to check the versions of Xcode being used on the supported versions of Apple stuff. @DominusExult can you comment on this?`
I'm compiling with current Xcode which runs Apple Clang 15 but I'm targeting macOS 10.11 as lowest supported macOS. If I compile with -std=c++17 and test on macOS 10.11 this should tell me whether the libc++ included runs ok, right?
There is a handy compiler support overview at https://en.cppreference.com/w/cpp/compiler_support/17

pathfinder/PathFinder.h
marzojr: The issue above can be fixed by applying the change I suggested to `pathfinder/Pathfinder.h`, and making the following change to schedule.cc:

```diff
-		Actor_pathfinder_client COST = (GAME_SI ? cost : cost2);
+		Actor_pathfinder_client& COST = (GAME_SI ? cost : cost2);
```

I did some looking up, and there is some const-correctness changes to pathfinder stuff that allows all of these pathfinder cost classes to be passed down as const I will be doing these changes myself, so you don't have to worry.

marzojr:
> I'm compiling with current Xcode which runs Apple Clang 15 but I'm targeting macOS 10.11 as lowest supported macOS.

Apple Clang 15 is more than enough, and better than the version on Android NDK. So I will be pulling the plug and changing to c++17.

Dragon-Baroque:
> I would agree even jumping straight to C++ 20.

I am quite at loss in comparing between C++ 11, C++ 14, and C++ 17. But I am eagerly waiting for C++ 20 with the `std::format` family.

Dragon-Baroque:
@marzojr,

Thank you for your explanation.

PR #420, opened 2024-01-01
Normalize all (v)snprintf, Silence macOS over ICU

Dragon-Baroque:
Ensure that all `(v)snprintf` 

- Set their size parameter to always an identifier or the `sizeof` of the receiving buffer,
- Are not followed by a store of a caution 00 byte at the end of the buffer, since `(v)snprintf` warrants the trailing 00.

Silence warnings about NULL in ICU on macOS in `mapedit/studio.cc`.


marzojr:
Hm. With the move to c++17, I am taking a few steps to get rid of array_size. I would prefer to use sizeof for these cases.

Dragon-Baroque:
Fine, I alter this commit to use `sizeof` instead of `array_size` right away.

marzojr (approved):

PR #421, opened 2024-01-01
Silence C++ 20 warning

Dragon-Baroque:
@marzojr, will you accept the following PR :

I tried to compile ( GCC and CLang ) with C++ 20, and I got two new warnings :

1- `arithmetic between different enumeration types ‘Ucscript::Ucscript_ops’ and ‘Actor::Frames’ is deprecated [-Wdeprecated-enum-enum-conversion]` over `actors.cc` and `schedule.cc` about `(Ucscript::npc_frame + Actor::kneel_frame)` and similar.

2- only in GCC :
```
In function 'void Write1(uint8*&, uint8)',
    inlined from 'void Write2(Dest&, uint16) [with Dest = unsigned char*]' at ../files/utils.h:309:8,
    inlined from 'static std::unique_ptr<unsigned char []> Shape_frame::encode_rle(unsigned char*, int, int, int, int, int&)' at vgafile.cc:313:8:
../files/utils.h:296:16: warning: null pointer dereference [-Wnull-dereference]
  296 |         *out++ = val;
      |         ~~~~~~~^~~~~
```

I propose to fix the first warning with the current PR as 

- Expand Ucscript::Ucscriptops npc_frame with the Actor::Frame enum in `ucscriptop.h`
- Replace `(Ucscript::npc_frame + Actor::ready_frame)` and similar by `Ucscript::npc_ready_frame` and similar in `actors.cc` and `schedule.cc`.

For the second warning, could you advise on how to convince ( and check, if possible ) the GCC compiler that a pointer _is not null_ ?

### DIFF
--- a/Makefile.mingw
+++ b/Makefile.mingw
@@ -160,7 +160,6 @@ WARNINGS := \
 	-Wall \
 	-Wextra \
 	-Wpedantic \
-	-Wc++14-compat \
 	-Wc++17-compat \
 	-Wc++20-compat \
 	-Walloc-zero \
--- a/actors.cc
+++ b/actors.cc
@@ -3173,24 +3173,21 @@ void Actor::lay_down(bool die) {
 
 	set_action(nullptr);
 	auto* scr = new Usecode_script(this);
-	(*scr) << Ucscript::finish << (Ucscript::npc_frame + Actor::standing);
+	(*scr) << Ucscript::finish << Ucscript::npc_standing_frame;
 	if (GAME_SI && get_shapenum() == 832) {    // SI Frost serpent
 		// Frames are in reversed order. This results in a
 		// strangeanimation in the original, which we fix here.
-		(*scr) << (Ucscript::npc_frame + Actor::sleep_frame)
-			   << (Ucscript::npc_frame + Actor::kneel_frame)
-			   << (Ucscript::npc_frame + Actor::bow_frame)
-			   << (Ucscript::npc_frame + Actor::sit_frame);
+		(*scr) << Ucscript::npc_sleep_frame << Ucscript::npc_kneel_frame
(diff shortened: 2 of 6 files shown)


marzojr (changes requested):
Make sure to update from master and run clang-format on the modified files.

marzojr:
The second one is a false-positive. I have been trying to convince GCC that it is a false-positive, but with no success so far.

Dragon-Baroque:
> Make sure to update from master.

Done.

> Run clang-format on the modified files.

Sure. Could you give me the right command ?

marzojr:
> **After try :** Just `clang-format -i <targetfile>`, is this correct ?

Yes. It should pick up the .clang-format file from the root of the repository and format with it.

Dragon-Baroque:
1. I compile in DEBUG and this fails because of COUT / CERR redefinition : in `exult_constants.h` and `headers/common_types.h`. I can alter the current PR with the removal you tell me to choose.
2. I have lot of _prior to C++ 17 warnings_, I suppressed them in removing `-Wc++14-compat`, I can alter also the current PR so.

marzojr:
> 1. I compile in DEBUG and this fails because of COUT / CERR redefinition : in `exult_constants.h` and `headers/common_types.h`. I can alter the current PR with the removal you tell me to choose.

Where does it fail? Can you post the error messages?

>     2. I have lot of _prior to C++ 17 warnings_, I suppressed them in removing `-Wc++14-compat`, I can alter also the current PR so.

Please do.

Dragon-Baroque:
Like it warns in ci-linux, except that I get an error instead of a warning.

marzojr:
Hm. Ah, I see. Let me handle that one.

marzojr (approved):

Issue #425, opened 2024-01-08
Review of m4/ directory and the files autogen.sh and configure.ac


Dragon-Baroque:
Proposal :

- Remove the content of the `m4/` directory and rely on the package `autoconf-archive` ( Linux and FreeBSD ) to provide the recent macros,
- Except for the `GB_ENABLE_LTO` ( Link Time Optimization ) which is not in `autoconf-archives`, which is for GCC only, that adds a `-flto` to `CFLAGS` when `--enable-lto` is passed to `./configure` ( we could embed it into `configure.ac` or leave it alone in `m4/` ), it builds a working Exult properly, but with ( apparently unrelated ) warnings, 
```
usecode/useval.cc:503:66: warning: argument 1 value '18446744073709551615' exceeds maximum object size 9223372036854775807 [-Walloc-size-larger-than=]
  503 |                 clsrefval.elems = new Usecode_value[clsrefval.cnt];
      |                                                                  ^
usecode/useval.cc:528:61: warning: argument 1 value '18446744073709551615' exceeds maximum object size 9223372036854775807 [-Walloc-size-larger-than=]
  528 |         clsrefval.elems    = new Usecode_value[clsrefval.cnt];
      |                                                             ^
shapes/font.cc:105:62: warning: argument 1 value '18446744073709551615' exceeds maximum object size 9223372036854775807 [-Walloc-size-larger-than=]
  105 |         auto*       lines          = new string[max_lines + 1];
      |                                                              ^
gumps/Newfile_gump.cc:1054:39: warning: argument 1 value '18446744073709551615' exceeds maximum object size 9223372036854775807 [-Walloc-size-larger-than=]
 1054 |         games = new SaveInfo[num_games];
      |                                       ^
gumps/Sign_gump.cc:64:42: warning: argument 1 value '18446744073709551615' exceeds maximum object size 9223372036854775807 [-Walloc-size-larger-than=]
   64 |         lines = new std::string[num_lines];
      |                                          ^
``` 
- On the other hand, if we empty `m4/` completely, then we must create a `dummy` marker file into it to prevent Git from removing it, or accept a warning by `aclocal` that `m4/` is missing,
- Remove the `autogen.sh` and replace it by a `autoreconf -fvi` in the GitHub workflows, like FreeBSD does,
- Reorder some log macros, by now we get :
```
checking for zipped savegame support... checking for zlib.h... yes
yes
```
instead of
```
checking for zlib.h... yes
checking for zipped savegame support... yes
```

PR #426, opened 2024-01-08
Cleanup configure.ac, autogen.sh, m4/ directory

Dragon-Baroque:
Per issue #425, 

- Remove macros in `m4/ax....m4`, provide them with `autoconf-archive`,
- Remove `autogen.sh`,  use `autoreconf -fvi` instead,
- Ensure output of `configure.ac` is not interlaced, 
- Install `autoconf-archive` and use `autoreconf` instead of `autogen.sh` in `.github/workflows/ci-android + linux + macos.yml, codeql.yml, smapshots-android.yml`,
- Alter an internal `CMakeLists.txt.in` for Android
- Review the `READMEs` and `INSTALL` that mention `autogen.sh` to advise for `autoreconf -fvi` instead, and add `autoconf-archive` to the prerequisites


DominusExult:
> Still struggling to get the `ci-android` workflow right. Making the PR a draft.

The Android build *needs* to be done outside of the code. Your change to `android/app/src/main/cpp/dependencies/exult/CMakeLists.txt.in`:
` CONFIGURE_COMMAND cd <SOURCE_DIR> && autoreconf -f -v -i`
probably makes it fail.

Dragon-Baroque:
Could you enlighten me ? 
At the point of failure, since this is out of tree, I expect builddir != sourcedir, but are what builddir and sourcedir ?

There are comments on the net, claiming that `autoreconf` does not like out of tree builds. On the other hand, `autogen.sh` runs the same autotools than `autoreconf.` Except, I suppose, `autogen.sh` uses its own location to find the source `configure.ac` and produces the result in the build directory, is that correct ?

DominusExult:
> Could you enlighten me ? At the point of failure, since this is out of tree, I expect builddir != sourcedir, but are what builddir and sourcedir ?
> 
> There are comments on the net, claiming that `autoreconf` does not like out of tree builds. On the other hand, `autogen.sh` runs the same autotools than `autoreconf.` Except, I suppose, `autogen.sh` uses its own location to find the source `configure.ac` and produces the result in the build directory, is that correct ?

I'm not exactly sure what happens. Quickly testing, it doesn't matter where you run autogen.sh from, it does create the files in the source folder, not any build folder. 
I can only say that the building fails after the configure step in `android/app/src/main/cpp/dependencies/exult/CMakeLists.txt.in` - For comparison see https://github.com/exult/exult/actions/runs/7446410361/job/20256602425#step:8:10807 from when it worked with your earlier GTK changes.

Dragon-Baroque:
> Quickly testing, it doesn't matter where you run autogen.sh from, it does create the files in the source folder, not any build folder.

And the result of `autoreconf` is actually most similar from the result of `autogen.sh`.

Now it seems that it depends very much how you code the call to `configure`.
It is this one that decide where to find the sources ( near where it is ), and where to build the `Makefile` ( what is the current directory when it is called ).
So my code `cd <SOURCE_DIR> && autoreconf -f -v -i` could be missing a change directory back, and could have left the current directory at the wrong place, so that `configure` was called from the wrong directory, ( source instead of build ), and produced the `Makefile` at the wrong place ( source instead of build ). So I would be quite close actually ...

Now I would like to stop polluting the Actions with my various tests. Could you tell me how I can reproduce the android build at home on my Linux ? Just follow the `ci-android.yml` ?

marzojr:
`ci-android.yml` runs on a Ubuntu VM. You can try setting up one, or reproducing the behavior locally if you are using Linux.

A possible alternative is to try out [act](https://github.com/nektos/act); it is not perfect, but should be good enough.

DominusExult:
> Now I would like to stop polluting the Actions with my various tests. Could you tell me how I can reproduce the android build at home on my Linux ? Just follow the `ci-android.yml` ?

https://github.com/exult/exult/tree/master/android#how-to-compile-for-android <- this worked for me. Setting it up was a bit of a work, though. But using a fork of Exult, disabling the notify options and then just let the CI's run is also a good way :)

Dragon-Baroque:
> Then just let the CI's run is also a good way.

I hoped to do that some time past. Just push a commit onto my fork, not a PR, and run just one CI. I could not make any CI run.
I just assumed that only a PR would trigger the CIs, and then all of them. Might work only from the root tree, not from a fork.

DominusExult:
I found this could be of relevance
https://answers.ros.org/question/333125/how-to-include-external-automakeautoconf-projects-into-ament_cmake/

```
CONFIGURE_COMMAND touch <SOURCE_DIR>/ChangeLog 
COMMAND autoreconf -i <SOURCE_DIR> 
COMMAND <SOURCE_DIR>/configure 
```
Seems CONFIGURE_COMMAND needs to be *something* and then you run autoreconf as a `COMMAND`

DominusExult:
Unfortunately I can't test on my machine right now as my Android stuff is broken again :(

DominusExult:
@Dragon-Baroque in my fork of your fork I was able to run the CI with the above change, see https://github.com/DominusExult/exult/actions/runs/7452987549/job/20277411342 (I had to do another small change to the ci-android.yml file to trigger the action, though).
https://github.com/DominusExult/exult/commit/c29ca9334252a35584e2e3037ac6c874fed7735f (ignore the ci-android.yml changes - just there to avoid a failure on the notify messages)

Dragon-Baroque:
@DominusExult, so simple, thank you, now the CIs all pass.

> Seems CONFIGURE_COMMAND needs to be something and then you run `autoreconf` as a COMMAND.

I caught this later : the various `CONFIGURE_COMMAND`, `BUILD_COMMAND`, etc, of the `CMakeLists.txt.in` can contain one or several OS commands, and then there are several, they are separated by the `COMMAND` keyword. In other words, the `CONFIGURE_COMMAND` of Android for exult, being :
```
  CONFIGURE_COMMAND autoreconf -f -v -i <SOURCE_DIR>
  COMMAND           . @ENVFILE@ && <SOURCE_DIR>/configure --prefix=<INSTALL_DIR> --host=@ANDROID_LLVM_TRIPLE@ --enable-libexult --enable-shared --disable-data --disable-tools --enable-zip-support --enable-midi-sfx --enable-mt32emu --disable-fluidsynth --disable-alsa --disable-timidity-midi --disable-oggtest --disable-vorbistest
```
is actually made of two OS commands, one `autoreconf` and one `configure`.
We might consider to indent the keyword `COMMAND` to make visible that it is a continuation of the `CONFIGURE_COMMAND`.

@marzojr, @DominusExult, thank you for the suggestions.

Dragon-Baroque:
I have reviewed also the `INSTALL` and the `READMEs`,
I have indented the `CMakeLists.txt.in` as I proposed :
```
CONFIGURE_COMMAND autoreconf -f -v -i <SOURCE_DIR>
          COMMAND . @ENVFILE@ && <SOURCE_DIR>/configure --prefix=<INSTALL_DIR> --host=@ANDROID_LLVM_TRIPLE@ --enable-libexult --enable-shared --disable-data --disable-tools --enable-zip-support --enable-midi-sfx --enable-mt32emu --disable-fluidsynth --disable-alsa --disable-timidity-midi --disable-oggtest --disable-vorbistest
```

All the CIs still pass.
Reverting to a normal ( ie non-draft ) PR.

DominusExult:
I think this is ready to go live. I have just one nitpick: in the github actions you install the `autoconf-archive` right before running autoreconf instead of the `Install dependencies`, which is fine for testing but should be in that step. And for consistency sake, the ci-android and android snasphsot file should have that extra step `Install dependencies` as well.

marzojr:
> I think this is ready to go live. I have just one nitpick: in the github actions you install the `autoconf-archive` right before running autoreconf instead of the `Install dependencies`, which is fine for testing but should be in that step. And for consistency sake, the ci-android and android snasphsot file should have that extra step `Install dependencies` as well.

I agree with this.

.github/workflows/ci-linux.yml
```diff
@@ -130,10 +130,10 @@ jobs:
             libgdk-pixbuf2.0-dev libxml2-dev bison flex timidity libgimp2.0-dev libicu-dev
       - name: Checkout code
         uses: actions/checkout@master
-      - name: Run autogen
+      - name: Run autoreconf
         run: |
-          chmod a+x ./autogen.sh
-          ./autogen.sh
+          sudo apt-fast install -y autoconf-archive
```
marzojr, line 135: This should be in `Install dependencies` step.

.github/workflows/ci-macos.yml
```diff
@@ -109,10 +109,10 @@ jobs:
           brew install sdl2 icu4c mt32emu
       - name: Checkout code
         uses: actions/checkout@master
-      - name: Run autogen
+      - name: Run autoreconf
         run: |
-          chmod a+x ./autogen.sh
-          ./autogen.sh
+          brew install autoconf-archive
```
marzojr, line 114: This should be in `Install dependencies` step.

.github/workflows/codeql.yml
```diff
@@ -109,10 +109,10 @@ jobs:
           sudo apt-fast install -y \
             zlib1g-dev libogg-dev libvorbis-dev libasound2-dev libfluidsynth-dev libsdl2-dev libpng-dev libfreetype6-dev libgtk2.0-dev libgtk-3-dev \
             libgdk-pixbuf2.0-dev libxml2-dev bison flex timidity libgimp2.0-dev
-      - name: Run autogen
+      - name: Run autoreconf
         run: |
-          chmod a+x ./autogen.sh
-          ./autogen.sh
+          sudo apt-fast install -y autoconf-archive
```
marzojr, line 114: This should be in `Install dependencies` step.

.github/workflows/coverity-scan.yml
```diff
@@ -47,11 +47,11 @@ jobs:
       - name: Checkout code
         if: ${{ env.EXULT_REPO_ALIVE == 'true' }}
         uses: actions/checkout@master
-      - name: Run autogen
+      - name: Run autoreconf
         if: ${{ env.EXULT_REPO_ALIVE == 'true' }}
         run: |
-          chmod a+x ./autogen.sh
-          ./autogen.sh
+          sudo apt-fast install -y autoconf-archive
```
marzojr, line 53: This should be in `Install dependencies` step.

marzojr (changes requested):

PR #429, opened 2024-01-14
SDL event wrapper refactor

### LINKED ISSUE #62: Normalization of SDL User Events in Exult
Exult uses SDL User Events inconsistently :
ShortcutBar_gump uses SDL_USEREVENT directly, TouchUI registers one User Event to SDL.
They end up using the same User Event.

I have published a commit in the branch "exult_user_events" on Dragon-Baroque/exult to 
centralize the management of SDL User Events.
A longer explanation is in the Exult_Review_of_SDL_User_Events.md.

marzojr:
This is an in-progress rework I am working on for wrapping SDL events and unifying input code and event loops in Exult. The new event manager is mostly done, I still need to hook things up everywhere. I also want to unify double-click handling so that shortcut bar uses the same code as everything else, as well as adding real support for gamepads (including button remapping), and adding support for the default game codepage in text input for the versions that support them (such as French BG) so that you can enter "à" in the avatar's name, for example.

At this stage, this pull request is mainly to see how many compilers/OSes I will be breaking with the code I have written.

My ideas for default gamepad bindings:
- left thumbstick makes you walk;
- right thumbstick moves the mouse pointer;
- left shoulder button is bound to inventory (`I` on keyboard);
- right shoulder button is bound to stats (`Z` on keyboard);
- back is mapped to `esc`;
- guide is bound to help (`H` on keyboard, maybe a specific version for gamepads);
- menu shows save menu (`S` on keyboard);
- A is bound to left click ("inspect");
- B is bound to double-click with left button ("interact");
- X is bound to toggle combat (`C` on keyboard);
- Y starts target mode (`T` on keyboard);
- d-pad up shows spellbook (`B` on keyboard);
- d-pad down shows map (`M` on keyboard);
- d-pad left and d-pad right: no idea
- left and right triggers: maybe act like mouse-wheel?
- right thumbstick button: maybe try to grab item under mouse cursor? And press it again to release?

Edit: TO-DO list:
- [ ] Unify double-click handling
- [ ] Add support for converting unicode input to stock game's codepage when appropriate
- [ ] Full gamepad support
- [ ] Hook up the event manager throughout the codebase
- [ ] Help @Dragon-Baroque update his fork (again) due to the massive changes this will entail


DominusExult:
Ah, thank you! I'm so looking forward to this.
Your mappings make sense, except I'd swap back and right shoulder. ESC is used way more often should be more comfortable to reach, and the back button is a bit of a hassle.
left and right triggers for mousewheel, sounds good, as well as the right thumbstick button idea.

marzojr:
@Dragon-Baroque I'd also like your input, especially since your SDL3 fork will be impacted by this.

DominusExult:
For SDL3 it might help with the issue I have on iOS when I use other fill modes than fill. Slouken,Sam Lantinga, from the SDL devs suggested using `SDL_ConvertEventToRenderCoordinates(SDL_Renderer *renderer, SDL_Event *event);` (https://github.com/libsdl-org/SDL/blob/483155bbf94b661b9054d1dba75b23ec9ed5626a/include/SDL3/SDL_render.h#L1139). Looking a bit into it, this is used mostly by getting the renderer and while polling for events set this on each found event.

Dragon-Baroque:
> Help **Dragon-Baroque** update his SDL3 fork (again) due to the massive changes this will entail.

At this point, the only _significant_ changes are in the new `eventman.h + cc`.

When you reach completion, where do you expect the _massive changes_ ?

- I suppose that `exult.cc`, plus a few places in the Gumps will have a whole replacement of their Event Handlers by a set of register callbacks, or shall there be more ?
- I also wonder by what you will replace the two existing `SDL_PumpEvents`.

So I have made a very quick attempt to compile `eventman.cc` against SDL 3. I just imported the `configure.ac` from the SDL3 branch, and I applied the SDL provided migration scripts. Nothing more. Then I compiled `eventman.cc` ( `make eventman.o` ) and I got few errors, all because of some API changes that the migration scripts could not handle and that I handled easily during my port :

- Mouse positions given in `float`, possibly for HighDPI ( this was actually the most cumbersome, because of its wide reach, until we agreed that Exult had no need for subpixel precision ),
- Joystick device enumeration being slightly different and data field being renamed,
- Window events being no longer subclassed,
- DragDrop events data field being renamed + Enablement API being slightly different.

If you consider `eventman.cc` to be close to stable, I may look seriously at the SDL3 required changes into it from the list above, to be more confident.

Dragon-Baroque:
Yes, I too wondered whether we should apply `SDL_ConvertEventToRenderCoordinates` to all events.
Having a center place where that would be done, and even better, be done ( in SDL 3 ) while we still hold the mouse positions in `float`, would mightily simplify the final event handlers. We would need to pass the `SDL_Renderer` to the `EventManager`.

Dragon-Baroque:
Finally, Sam Lantinga pointed out that SDL 3 allows a new application architecture where the event loop is no longer in the application code, and where the application provides four callbacks, the two most important of which are :

- `SDL_AppIterate( void )` called at a regular pace, the heart beat of the application,
- `SDL_AppEvent( const SDL_Event *event )` called on a SDL event.

Do you think you could take advantage of that ?

By the way, I noticed that those callbacks do not offer a reentrancy anchor parameter, like the last parameter named userdata of all GTK callbacks, and I wonder whether the current `Event Manager` provides for one, or even if the Exult callbacks need one.

marzojr:
> At this point, the only _significant_ changes are in the new `eventman.h + cc`.

It is, indeed, because I have been writing it so that it can replace all the other event loops in Exult. It is now good enough that I will start doing exactly that. So:

> When you reach completion, where do you expect the _massive changes_ ?

Basically, everywhere that we get SDL events and act based on them, as well as most places where SDL headers are included. The former will conflict with every change you made in your fork to those event loops to get them to compile and work with SDL3. And since I plan to confine all SDL event stuff to be inside `eventman.cc` (not even in the header, I was just a bit lazy so far with SDL_KeyCode and SDL_Keymod), as well as every place that currently uses SDL events, or the data in them.

>     * I also wonder by what you will replace the two existing `SDL_PumpEvents`.

Hm, I had missed the SDL_PumpEvents + SDL_PeepEvents places. It should be a simple matter to add a method that does basically that, though; I am aware of the place using SDL_WaitEvent (ignoring `soundtest.cc`), but this one can be easily rewritten in terms of SDL_PollEvent.

>     * Mouse positions given in `float`, possibly for HighDPI ( this was actually the most cumbersome, because of its wide reach, until we agreed that Exult had no need for subpixel precision ),

These should be a simple fix; I saw what you did with those, and it should be a matter of porting those. Oh, and I forgot to mention: just doing static_cast to int from float is probably not ideal, we should probably do a `round` or `floor` before.

> If you consider `eventman.cc` to be close to stable, I may look seriously at the SDL3 required changes into it from the list above, to be more confident.

It is mostly stable; I will be rewriting some things for the keyboard events (to wrap around SDL_KeyCode/SDL_Keymod), implementing the controller stuff, implementing proper double-clicks (so shortcut bar and in-game double-clicks work the same way), and moving remaining touch-UI stuff into `eventman.cc`; after that, I don't expect to need to do much more on it other than assist in porting to SDL3.

> Finally, Sam Lantinga pointed out that SDL 3 allows a new application architecture where the event loop is no longer in the application code, and where the application provides four callbacks, the two most important of which are :
[snip]
> Do you think you could take advantage of that ?

It should be doable, yes.

> By the way, I noticed that those callbacks do not offer a reentrancy anchor parameter, like the last parameter named userdata of all GTK callbacks, and I wonder whether the current `Event Manager` provides for one, or even if the Exult callbacks need one.

That is incorrect; but it is more a documentation issue than anything, since the code in this part is very hard to follow if you are not a c++ expert (with the intention of being as easy to use as possible).

`EventManager::register_one_callback` has 3 forms:
1. One form that basically takes function pointers;
2. One form that basically takes functor and lambdas;
3. One form that takes a "data" parameter and any callable thing that can use this "data" parameter, along with the callback's normal parameters, to make the callback call.

`EventManager::register_callbacks` has 2 forms:
1. takes many callable things, calls `EventManager::register_one_callback` for each of them, and aggregates the results;
2. takes one "data" parameter and many callable things, calls `EventManager::register_one_callback` for "data" plus each of the callables, and aggregates the results.

All of those return RAII-types that will remove the callbacks from the stack when they are destroyed.

So not only can you pass a "data" parameter to the callbacks, but in some cases (functor and labmdas) they will be inferred without needing to be specified, and in call cases they will be done in a type-safe way that will cause a compilation error if the types are wrong -- so no more casting from a `void*` data parameter.

marzojr:
@wench Any idea what could be causing this on msvc build:
>       4>eventman.h.obj : error LNK2001: unresolved external symbol __imp_MapViewOfFileNuma2 [D:\a\exult\exult\msvcstuff\vs2019\Exult.vcxproj]
>     4>D:\a\exult\exult\msvcstuff\vs2019\..\..\Exult.exe : fatal error LNK1120: 1 unresolved externals [D:\a\exult\exult\msvcstuff\vs2019\Exult.vcxproj]

Or better yet, what can I do to fix it?

wench:
No idea what would be causing that and why its trying to compiles the header as an an object guessing that is part of link time code generation. but it seems like the link error would probably be fixed by retargeting the Exult project to use a newer Windows 10 SDK version (1703 build 15063) as the missing function was added in Windows 10 1703 and I am building Exult against the SDK for the original release version of Windows 10. I am curious what in that header would even need to call that function

wench:
Though it does not look like the github runner has the 1703 build 5063 sdk installed so you'd need to target one they do have like 1709 build 16299. I think it should be able to just edit the vcxproj and change the  WindowsTargetPlatformVersion line from 10.0 to 10.0.16299.0

marzojr:
I copied the error and had completely missed the fact that it was trying to compile the header as an object file. I guess I was focusing too much at the error message. Anyway, trying now with a dumb error fixed; hopefully, it will auto-fix the issue.

wench:
Nice to see it was fixed so simply. Copy and Paste error?

marzojr:
> Nice to see it was fixed so simply. Copy and Paste error?

Yup. That should teaches me not to edit these files by hand 😁

marzojr:
So I was looking into rebasing this to master... and the changes to cheat screen basically killed my motivation for now. I will see about restarting thew work by first consolidating all event manager stuff, and adding support for the new stuff added in cheat screen; then start pulling in the changes hooking the event manager and redoing the work as needed.

DominusExult:
I noticed that the move to SDL2 killed "special" characters in text input. Writing punctuation or any character (other than uppercase letters) that is input by shift+key are not written. This is especially annoying in the Notebook gump and only slightly useful for the Newfile Gump.
Could this be fixed by this PR?

Issue #436, opened 2024-01-28
Fails to build from 20240110 -> 20240118: ld: error: undefined symbol: pthread_create


nunotexbsd:
FreeBSD clang version 17.0.6

```
/bin/sh ./libtool  --tag=CXX   --mode=link c++ -std=c++17  -O2 -pipe -fstack-protector-strong -isystem /usr/local/include -fno-strict-aliasing -isystem /usr/local/include -I/usr/local/include   -fstack-protector-strong  -o exult exult-actions.o exult-actorio.o exult-actors.o exult-args.o exult-browser.o exult-cheat.o exult-cheat_screen.o exult-combat.o exult-dir.o exult-drag.o exult-effects.o exult-exult.o exult-exultmenu.o exult-game.o exult-gameclk.o exult-gamedat.o exult-gamemap.o exult-gamerend.o exult-gamewin.o exult-istring.o exult-keys.o exult-keyactions.o exult-menulist.o exult-monsters.o exult-mouse.o exult-npcnear.o exult-npctime.o exult-palette.o exult-party.o exult-paths.o exult-readnpcs.o exult-schedule.o exult-shapeid.o exult-touchui.o exult-tqueue.o exult-txtscroll.o exult-verify.o exult-version.o pathfinder/libpathfinder.la audio/libaudio.la audio/midi_drivers/libmididrv.la audio/midi_drivers/timidity/libtimidity.la conf/libconf.la imagewin/libimagewin.la server/libserver.la gamemgr/libgamemgr.la flic/libflic.la usecode/libusecode.la shapes/libshapes.la shapes/shapeinf/libshapeinf.la objs/libobjs.la gumps/libgumps.la files/libu7file.la files/sha1/libsha1.la files/zip/libminizip.la  -lpng -L/usr/local/lib -lSDL2  -L/usr/local/lib -lX11 -lXext  -lz   -L/usr/local/lib -logg -lvorbis -lvorbisfile    -L/usr/local/lib
[...]

PR #437, opened 2024-01-31
Rework configure.ac for PThread, comment OS choices

Dragon-Baroque:
Normalize all POSIX platforms using GCC or CLang to use `-pthread`,
Comment a bit the support state of the various platforms.

Proposition to #436.


marzojr (approved):
As I mentioned on the issue page, it probably should go to OpenBSD, NetBSD, and Solaris as well.

marzojr (approved):

PR #440, opened 2024-02-02
Build Exult for FreeBSD

Dragon-Baroque:
A first attempt to build Exult for FreeBSD


marzojr:
Out of curiosity: if something goes *wrong* in the `prepare` or `run` steps, does it accuse as such?

Otherwise, it looks good to me; but you can copy/edit the IRC/Discord notification setup from the other CIs, and add this to our `README.md`:

```markdown
[![GitHub Actions Workflow Status](https://img.shields.io/github/actions/workflow/status/exult/exult/ci-freebsd.yml?logo=freebsd&label=CI%20FreeBSD&labelColor=white)](https://github.com/exult/exult/actions/workflows/ci-freebsd.yml)
```

Dragon-Baroque:
> Out of curiosity: if something goes wrong in the prepare or run steps, does it accuse as such?

From the code, I would say yes onto the `run` step. The `prepare` step seems much less tracked.
In the other workflows, what is it that triggers a job fail ? A non zero error level by some step ? I can try to simulate that with an `exit 1` or so in the two steps. Would that help ?

Now another question : my first run did not involve the Gnome Thumbnailer because on my VM the `sed` step supposed to build `gnome_u7shapes.thumbnailer` froze. In debugging it looked as if the input and output redirections to `sed` where set wrong. I fixed that by replacing `make` by `gmake`. Does it mean anything to you ?

Dragon-Baroque:
With an `exit 1` in `prepare` : the job stops and is marked failed.

marzojr:
> From the code, I would say yes onto the `run` step. The `prepare` step seems much less tracked. In the other workflows, what is it that triggers a job fail ? A non zero error level by some step ? I can try to simulate that with an `exit 1` or so in the two steps. Would that help ?

Yes, a job fails on the other workflows when a command returns a nonzero value. Even if `prepare` can't be made to fail in this way, it is still useful to know that it can fail on `run` -- we can add it to CI and get notifications of a broken build.

> Now another question : my first run did not involve the Gnome Thumbnailer because on my VM the `sed` step supposed to build `gnome_u7shapes.thumbnailer` froze. In debugging it looked as if the input and output redirections to `sed` where set wrong. I fixed that by replacing `make` by `gmake`. Does it mean anything to you ?

I assume you mean this:
```makefile
$(thumbnailer_DATA): $(thumbnailer_in_files)
	@$(SED) -e "s|\@bindir\@|$(bindir)|" $< > $@
```
By default, BSD variants use BSD `make` instead of GNU `make`. When you call `make` in BSD, you get BSD `make`; and when you call `gmake` you get GNU `make`. Since BSD shell is also `tcsh`, I assume that it is what BSD `make` uses, whereas using GNU `make` gets a variant of `sh`.

But why it is hanging, I am unsure. Can you run `make` (not `gmake`) with `-n` and print what the `sed` command is? This can help us understand what is going wrong.

Dragon-Baroque:
> Yes, a job fails on the other workflows when a command returns a nonzero value. Even if prepare can't be made to fail in this way, it is still useful to know that it can fail on run -- we can add it to CI and get notifications of a broken build.

Please see my ( many times updated ) comment above. Both `prepare` and `run` can trigger a job failure, `run` is simply more verbose in case of an error to help fix it.

Run `make -n` :

```
rm -f gnome_shp_thumbnailer
/bin/sh ../libtool  --tag=CXX    --mode=link clang++ -std=c++17   -I/usr/local/include -isystem /usr/local/include -pthread -I/usr/local/include -pthread   -o gnome_shp_thumbnailer gnome_shp_thumbnailer.o ../shapes/libshapes.la  ../imagewin/libimagewin.la  ../files/libu7file.la  -lpng -lz  -L/usr/local/lib -lgdk_pixbuf-2.0 -lgobject-2.0 -lglib-2.0 -lintl -L/usr/local/lib
/usr/bin/sed -e "s|\@bindir\@|/usr/local/bin|"  > gnome_u7shapes.thumbnailer
```
As you can see, no __input__ redirection.

Run `make -dA`, end of log :
```
Examining x-shapefile.xml...   Caching 16:02:11 Feb 01, 2024 for x-shapefile.xml
modified 16:02:11 Feb 01, 2024...up-to-date.
Examining gnome_u7shapes.thumbnailer.in...   Caching 16:02:11 Feb 01, 2024 for gnome_u7shapes.thumbnailer.in
modified 16:02:11 Feb 01, 2024...up-to-date.
Examining gnome_u7shapes.thumbnailer...Wildcard expanding "gnome_u7shapes.thumbnailer"...
Searching for gnome_u7shapes.thumbnailer ...
   failed.
Found 'gnome_u7shapes.thumbnailer' as '(not found)'
nonexistent...modified before source "gnome_u7shapes.thumbnailer.in"...out-of-date.
gnome_u7shapes.thumbnailer: > = gnome_u7shapes.thumbnailer.in
gnome_u7shapes.thumbnailer: ? = gnome_u7shapes.thumbnailer.in
Var_Parse: $(SED) -e "s|\@bindir\@|$(bindir)|" $< > $@ (eval)
Var_Parse: $(bindir)|" $< > $@ (eval)
Var_Parse: ${exec_prefix}/bin (eval)
Var_Parse: ${prefix} (eval)
Var_Parse: $< > $@ (eval)
Var_Parse: $@ (eval)
/usr/bin/sed -e "s|\@bindir\@|/usr/local/bin|"  > gnome_u7shapes.thumbnailer
Execute: '/usr/bin/sed -e "s|\@bindir\@|/usr/local/bin|"  > gnome_u7shapes.thumbnailer'
Var_Parse: ${.MAKE.EXPORTED:O:u} (eval)
Evaluating modifier ${.MAKE.EXPORTED:O} on value "META_MODE"
Result of ${.MAKE.EXPORTED:O} is "META_MODE"
Evaluating modifier ${.MAKE.EXPORTED:u} on value "META_MODE"
Result of ${.MAKE.EXPORTED:u} is "META_MODE"
+ /usr/bin/sed -e 's|\@bindir\@|/usr/local/bin|'
<--- Ctl-C by me
make: *** gnome_u7shapes.thumbnailer removed
*** Signal 2
```

Is `$<` valid ?

marzojr:
> Please see my ( many times updated ) comment above. Both `prepare` and `run` can trigger a job failure, `run` is simply more verbose in case of an error to help fix it.

Excellent!

> ```
> /usr/bin/sed -e "s|\@bindir\@|/usr/local/bin|"  > gnome_u7shapes.thumbnailer
> ```
> 
> As you can see, no **input** redirection or **input** file.

I was suspecting something like this might be happening.

> Is `$<` valid ? It looks like a GNU Make feature...

The [`man page` for FreeBSD `make` ](https://man.freebsd.org/cgi/man.cgi?make(1)) seems to imply it is valid:
>        The $@ and $< variables are more or less universally portable, as is
>        the $(MAKE) variable. Basic use of suffix rules (for files only in the
>        current directory, not trying to chain transformations together, etc.)
>        is also reasonably portable.

Does it work if you just replace the `$<` by `$(thumbnailer_in_files)` directly?

marzojr:
> > Does it work if you just replace the `$<` by `$(thumbnailer_in_files)` directly ?
> 
> From `make -n`, the answer is yes.

Then add that change in and swap back to BSD make. This will help in making sure our build is portable.

.github/workflows/ci-freebsd.yml
marzojr, line 1: Remember to remove the debug commands before pushing (from both `prepare` and `run`):
- the `echo` lines
- the `pkg-info` lines
- the lines checking the environment state (`pwd`, `ls -lah`, `whoami`, `env`, `freebsd-version`)

marzojr (changes requested):

.github/workflows/ci-freebsd.yml
Dragon-Baroque: Done, thank you.

At this point, the port to FreeBSD is working and it does not disturb the other builds.
But I would like to remove the `-isystem /usr/local/include` ( Vorbis and Alsa ) and the `-L/usr/local/lib` ( mainly LibPNG ).
For this I need some more changes in `configure.ac` and the various `Makefile.am`.
You wanted to play now with the *BSD etc builds, so I am asking whether I should UnDraft and Publish the PR as it is now, or whether you can wait a few days.

marzojr (approved):

marzojr:
> But I would like to remove the `-isystem /usr/local/include` ( Vorbis and Alsa ) and the `-L/usr/local/lib` ( mainly LibPNG ). For this I need some more changes in `configure.ac` and the various `Makefile.am`.

That would be nice, yes.

> You wanted to play now with the *BSD etc builds, so I am asking whether I should UnDraft and Publish the PR as it is now, or whether you can wait a few days.

You basically did what I wanted to do anyway, so you can undraft and publish.

One question: do you think it would be worth to set up CI for some of the other VMs provided by VM Actions? I am thinking primarily of the other BSDs (mainly OpenBSD, and NetBSD, but maybe DragonflyBSD?), but Solaris might also be interesting.

Dragon-Baroque:
> Do you think it would be worth to set up CI for some of the other VMs provided by VM Actions ?

YES. The rework about `configure.ac` and the `Makefile.am` is a nice to have for FreeBSD, but it is needed for NetBSD, for one.

But a quick and dirty attempt to build Exult against OmniOS has shown portability issues, on the other hand, as a `char` is identical to a `sint8` and that causes a failure in `files/utils.h` - replication of `Read1` and `Write1`.

Dragon-Baroque:
> Do you think it would be worth to set up CI for some of the other VMs provided by VM Actions ?

Now, you and @DominusExult tell me whether this is worth the effort.

DominusExult:
IMO these other VMs are so niche, that I think it's way over the top creating CIs for those

marzojr:
> IMO these other VMs are so niche, that I think it's way over the top creating CIs for those

By this metric, we would have to drop anything other than Android, Windows, MacOS and iOS [ref](https://gs.statcounter.com/os-market-share#monthly-202312-202312-bar)

DominusExult:
🤷‍♂️🤪

wench:
> > But I would like to remove the `-isystem /usr/local/include` ( Vorbis and Alsa ) and the `-L/usr/local/lib` ( mainly LibPNG ). For this I need some more changes in `configure.ac` and the various `Makefile.am`.
> 
> That would be nice, yes.
> 
> > You wanted to play now with the *BSD etc builds, so I am asking whether I should UnDraft and Publish the PR as it is now, or whether you can wait a few days.
> 
> You basically did what I wanted to do anyway, so you can undraft and publish.
> 
> One question: do you think it would be worth to set up CI for some of the other VMs provided by VM Actions? I am thinking primarily of the other BSDs (mainly OpenBSD, and NetBSD, but maybe DragonflyBSD?), but Solaris might also be interesting.

Solaris might be interesting but compiler and runtime support might be a problem as the Oracle Sun C++ compiler does not have any official C++17 support see: https://en.cppreference.com/w/cpp/compiler_support/17 We'd need to use gcc which makes it a little less interesting.

Dragon-Baroque:
> We'd need to use gcc which makes it a little less interesting.

Would you please elaborate on _why it makes it a little less interesting_ ?
The build process on Windows also uses GCC...

marzojr:
> Would you please elaborate on _why it makes it a little less interesting_ ?

Because it would have been one more compiler to test on if the Oracle Sun compiler supported C++17: there tends to be some divergence on implementations of C and C++, and knowing what breaks because one more compiler does things differently is useful.

> The build process on Windows also uses GCC...

This is mostly because I personally use the MinGW/msys2 builds, and develop using it while on Windows. I am not opposed to a MSVC build from the command line, but I don't have enough MSVC experience to do that.

wench:
I'll look into A MSVC workflow, from a casual look it should be fairly straightforward to setup a msbuild workflow

PR #442, opened 2024-02-05
Msvc workflow

wench:
Adding a workflow file to test build Exult with MSVC. Because this doesn't use makefiles like everything else, I expect that if anyone adds or removes source files this will fail and I'll need to fix it. No IRC/Discord Notification for now, I'll add that once pulled because it needs the project secret

### DIFF
--- a/.github/workflows/ci-msvc.yml
+++ b/.github/workflows/ci-msvc.yml
@@ -3,13 +3,60 @@
 # separate terms of service, privacy policy, and support
 # documentation.
 
-name: CI-MSVC
+name: ci-msvc
 
 on:
   push:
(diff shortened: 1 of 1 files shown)


Dragon-Baroque:
This workflow does not build Exult Studio, does it ?

wench:
No it does not. I haven't got Exult Studio building in MSVC. I'll look into it if there is a suitable nuget package for GTK

Dragon-Baroque:
>  I'll look into it if there is a suitable nuget package for GTK.

Studio : GTK+ 3 _and libpng_,
The Gimp plugin - much less important : GTK+ 2 and full GIMP.

PR #445, opened 2024-02-06
Build Exult on Solaris

Dragon-Baroque:
To Build Exult on Solaris derived OmniOS, because `sint8` is `char`, not `signed char` :

- `configure.ac` : Remove X11 from the SYSLIBS
- `files/utils.h` : Rework the `Read1(s)` and `Write1` functions using ((un)signed) char
- `tools/wuc.cc` : Replace the variable named `extended` by `is_extended`, reserved word


wench (approved):
looks good

Dragon-Baroque:
Not visible in the check list, but it passed also a build test on my OmniOS VM. That was the point of the patch.

PR #477, opened 2024-04-01
Moving to std::filesystem

marzojr:
This is a test to see how good support is in CI.


wench:
yeah, less platform specific code for the win. In would like to see some more of this, in particlar U7OpenDir in files/utils.cc that is actually using posix api calls/types and I need to kludge around it in the msvc build. This function is only used in one place to check the existence of a  directory and should be removed and replaced with a specific U7IsDir function that uses std::filesystem.

marzojr:
I am thinking of being slightly less conservative than this and just go for broke. get_system_path gets positively trivial with std::filesystem. U7open_in, U7open_out, U7opendir, and U7remove can all be properly implemented as case insensitive (by trying case-sensitive match first, then trying to find the latest parent directory which exists case-sensitively, then doing a case insensitive search for each path component that did not exist before, and **caching the resolve results**). And so on.

From what I see, the sole use of U7opendir (outside of U7exists in utils.cc) is in studio.cc for basically listing files in a directory and doing some globbing. So it could basically use U7ListFiles... even in master.

DominusExult:
With the updated Android build files (PR #512) you can now set ANDROID_NDK_VERSION=22.1.7171670 in configure.ac and it should just work.

marzojr:
So I have been working on this a bit. I have one absolutely giant commit which I am not sure I can make smaller because it is a mess of interconnected changes.

And I am thinking that maybe I want to work on adding cmake support first, with the intent of using [fmt](https://github.com/fmtlib/fmt), [scnlib](https://github.com/eliaskosunen/scnlib), and maybe [spdlog](https://github.com/gabime/spdlog/). They would all help improve some of the things I am working on on both this pull request, and on #534.

Another issue is the overlap of this and #534. Both make several similar changes on similar lines, and merging these is going to be a pain.

So I am thinking of creating a dev branch that (1) adds cmake support, (2) starts using fmtlib and scnlib in places where snprintf/scanf/etc are used, and (3) merge this pull and #534 onto said branch to work.

alagner:
(...)
> So I am thinking of creating a dev branch that (1) adds cmake support, (2) starts using fmtlib and scnlib in places where snprintf/scanf/etc are used, and (3) merge this pull and #534 onto said branch to work.

@marzojr I've started fiddling around with cmake some time ago, consider reusing [this](https://github.com/alagner/exult/tree/cmake). Unless of course you prefer a completely different cmake style

Issue #155, opened 2022-02-18
AntiMagic "Sparkle" weather is too long


DominusExult:
Continued from https://sourceforge.net/p/exult/bugs/1865/
In the original the antimagic weather changed sooner.
The problem is that the Sparkle egg has a duration of 0 which means it is supposed to be continuously. 
In https://github.com/exult/exult/blob/master/objs/egg.cc#L1097 we set the duration for "continuous weather" to 6000 which makes the sparkle weather stop_time extremely long.
This is a big problem when you step out of the cube generator as the antimagic rain just won't stop for ages. In the original it changed pretty soon (~10 seconds).

So either 
- we need to change the len(gth) for sparkle weather in egg.cc to something much smaller 
- or we do something wrong in the usecode events when the cube generator gets destroyed (maybe it stops the sparkle weather?)
- or we need a very small distance check from the sparkle weather egg

Cube generator sparkle weather egg is at 1854/2812

NotStanley4330:
I decided to do some digging on this. I tested out the cube weather effect in both Exult and DosBox. A few notes before I look any closer at code.

1) Exult seems to implement this weather effect all wrong. The effect should be a sparkly noise on screen as shown in the video below:


https://github.com/exult/exult/assets/44741162/f3c0c033-9552-4ce9-8ecd-f668c0272fc5

However in Exult it appears to be doing a thunderstorm:

https://github.com/exult/exult/assets/44741162/40323605-50d9-41d8-8fd4-5490eb9276c2

2) While in the cube, the sparkle effect does not seem to wear off at all. I haven't thoroughly tested but across several attempts it seemed to never go away and then consistently fade in a few seconds after leaving the cube. Thus it would appear to be something in the destruction of the cube generator and not in the egg itself. 

I can certainly dig at item number 2 a bit later today, but I have no clue where to begin on item number 1

NotStanley4330:
This video demonstrates what is happening in the cube in Exult (no sparkle):

https://github.com/exult/exult/assets/44741162/8c85ae83-e03e-45ad-a323-a0a2d221ad5e

And this one when you destroy the cube in DOSBox:


https://github.com/exult/exult/assets/44741162/3bcd3300-8032-4b24-bb14-1efa2dda87f3

NotStanley4330:
I did some testing to see what usecode is being executed when you double click on the Enclosed Sphere object to trigger the cube generator destruction, and here are some preliminary results.
[...]

PR #481, opened 2024-04-07
Issue #155 fix

NotStanley4330:
This PR is intended to fix some of the current issues with the "sparkle" storm effect that is present on the island of ambrosia and inside of the Cube Generator. The current issues are

1. Sparkle effect does not fade away after destroying the generator or leaving Ambrosia.
2. Sparkle effect does not work indoors, whereas in the original game it shows up inside the generator and in the cave outside of it after.
3. Sparkles effect looks different than in DOSBox, resembling rain instead of a general static noise of the original

For each of these problems we have identified the cause and have solutions:
1. This is caused by multiple weather effects being called by an egg but the egg not being copied on to the new effect caused the continuous effect not to be cancelled when the Avatar gets far enough away from the egg. I have made it so that the sparkle effect egg is passed on when it creates the new effect, but I can still add the functionality to more weather types.

2. This is due to the rain effect being programmed not to display indoors. The sparkle effect may be better suited to being it's own weather effect, but for now I just added an OR to make it so that it would display indoors if the weather type is sparkle.

3. This is due to being built on top of the rain effect, which has particles move and fall like rain. Fix still WIP.


NotStanley4330:
Opened pull request for review

marzojr (approved):

Issue #268, opened 2022-11-10
Single Sound Effect repeats on Energy Field


KnightCaptainU7:
The Energy Field shape, 768 in SI, is set to "Single Sound Effect" to match when it is first created. But the sound repeats over and over. I don't believe that is how it worked in the original game.

Steps to recreate:
Start a new game and teleport to Vasculio's area. Or drop one in via ES.

DominusExult:
Actually this behaves as the original. Simple steps to reproduce is via the energy field spell (no need to travel to Vasculio).
One could argue that this is annoying, though, but that's a whole world of grief when you go through all SFX.

KnightCaptainU7:
In testing via DOSbox, once the shape is in its final frame the related sound effect stops. In Exult, it just repeats over and over.

Is this something closed as wontfix? If so, I can disable the related sound in my mod.

DominusExult:
I've reopened it.
I think what happens and why I thought we were doing it correctly was that, as you said it stops when the energy field is complete but then starts again when the field goes down again.
And this happens when you cast the field but it doesn't go down when it is done by an egg.

DominusExult:
currently sfx are dehardcoded and editable in shape_info.txt (so in a mod you can turn that sfx off completely).
But the `%%section shape_sfx` does not have the option to limit the frames that play the sfx. That would need an extension or another `%%section shape_sfx` like section.
@marzojr what do you think?

@KnightCaptainU7 can you give me the exact coordinates (or rather close so I'm not on it) for an energy field to test in the original and Exult?

NotStanley4330:
I think I have this one fixed! I decided the easiest way was to extend the ```%%section shape_sfx``` section to have a field I have called halt_frame. Now the section is formatted as such:
```
:shapenum/sfxnum/delay/range/rand/extra/halt_frame
```

I added the necessary functions and lines to read this in and to check for it. Here is the new behavior of the energy field:


https://github.com/exult/exult/assets/44741162/43838d23-7980-4403-ae1a-ab39127da612

compared to how it currently behaves:

https://github.com/exult/exult/assets/44741162/e4670c0a-7bc3-46ec-ad1c-21a16a36365d

I don't think this will break anything, but if it does let me know. I created [this PR](https://github.com/exult/exult/pull/483) for the fix.

DominusExult:
That was fast! Thank you!

I am wondering something else but related since you dove into the sfx code.
[...]

PR #483, opened 2024-04-09
Energy field fix

NotStanley4330:
Fix for [#268](https://github.com/exult/exult/issues/268). This extends the functionality of ```sfxinc.h``` and the ```shape_info.txt``` file by adding a halt_frame to the possible fields in ```%%section shape_sfx``` . In order for this frame to be used and halt playing the sound effect I created a simple inline function to compare the shapes current frame against the listed halt frame. If the current frame is the halt frame, then the sfx should halt.


objs/animate.cc
```diff
@@ -59,6 +59,22 @@ static inline bool Get_sfx_out_of_range(
 		   > (MAX_SOUND_FALLOFF * MAX_SOUND_FALLOFF);
 }
 
+static inline bool Is_halt_frame(
```
marzojr, line 62: Please run clang-format on this file.

marzojr (changes requested):

marzojr:
Hm, see also my comment on the issue thread.

objs/animate.cc
NotStanley4330: I think I ran clang-format correctly. Please let me know if it didn't reformat correctly.

NotStanley4330:
I am going to close this as I have replaced it with an entirely different solution

PR #485, opened 2024-04-09
Enable sfx to have individual volume levels

NotStanley4330:
This PR will allow shape sfx to be individually adjustable via a field in ```%%section shape_sfx``` within ```shape_info.txt``` like so:
```
:shapenum/sfxnum/delay/range/rand/extra/volume
```

This PR currently has some values for a few of them, but getting to acceptable volume levels for each sfx should be a community effort.


NotStanley4330:
This PR also added the chance value to the energy field effect, helping it behave a bit more like the original (but still less annoying)

Issue #488, opened 2024-04-09
Main menu fps setting always shows 5fps


DominusExult:
Even after setting it to 8fps (or any other), after restart the main menu will show 5fps but when you start a game it will be the set fps value.

NotStanley4330:
Alright so I see that the issue is that we don't seem to actually pull in the info to the settings menu initially. I wonder why we do it this way:

```
const int realframes = 1000 / gwin->get_std_delay();

frames                        = -1;
framerates[num_default_rates] = realframes;
for (size_t i = 0; i < num_default_rates; i++) {
if (realframes == framerates[i]) {
	frames = i;
	break;
}
}

size_t num_framerates = num_default_rates;
if (frames == -1) {
num_framerates++;
frames = num_default_rates;
}
frametext.clear();
for (size_t i = 0; i < num_framerates; i++) {
frametext.emplace_back(framestring(framerates[i]));
}
```
So it creates this dynamically on startup and doesn't load in the frame values until you start a game, but do we have a distinct need for these things to be created dynamically? I can work within it however. Just seems odd.

DominusExult:
That's odd. It should just read it in the setting from the xml. But might be that historically it had some reason two decades ago :)

NotStanley4330:
> That's odd. It should just read it in the setting from the xml. But might be that historically it had some reason two decades ago :)

I agree. I can see that it's creating a list of text options but it just never actually loads from the xml. Should be like a 2 line fix

NotStanley4330:
Got a PR submitted for this one!

NotStanley4330:
New PR submitted that should fix this and rework the backend slightly to be less convoluted

PR #500, opened 2024-04-11
Main menu fps fix

NotStanley4330:
This is a fix for [Issue #488](https://github.com/exult/exult/issues/488) .  The fps button was not reading in a value from the exult.cfg file and instead simply defaulted to 5 on menu creation. Now it properly reads in the value and sets the button up accordingly.


DominusExult:
CodeFactor complains about extra line in 214. Could you fix that?

NotStanley4330:
> CodeFactor complains about extra line in 214. Could you fix that?

Fixed it!

gumps/GameEngineOptions_gump.cc
```diff
@@ -180,6 +182,35 @@ void GameEngineOptions_gump::load_settings() {
...
+		frames = 5;
+		break;
+	case 8:
+		frames = 4;
+		break;
+	case 5:
+		frames = 3;
+		break;
+	case 6:
+		frames = 2;
```
rofl0r, line 201: 6 has less fames than 5 ?
apart from that is seems the formula is frames = fps / 2 + fps % 2

NotStanley4330: So oddly enough this is how it orders it. The fps button cycles in this order
{2,4,6,5,8,10}
I was mostly seeking to not change what was there too much, but I can completely redo it to have it make more sense

NotStanley4330: this is the pre existing code

```
const int realframes = 1000 / gwin->get_std_delay();

	//this code generates the number of framerate options and their text
	//the functionality is kind of odd but I am leaving it alone.
	frames                        = -1;
	framerates[num_default_rates] = realframes;
	for (size_t i = 0; i < num_default_rates; i++) {
		if (realframes == framerates[i]) {
			frames = i;
			break;
		}
	}

	size_t num_framerates = num_default_rates;
	if (frames == -1) {
		num_framerates++;
		frames = num_default_rates;
	}
	frametext.clear();
	for (size_t i = 0; i < num_framerates; i++) {
		frametext.emplace_back(framestring(framerates[i]));
	}
```

NotStanley4330: To be clear the ```frames``` value really just tells us the index of the value in the vector ```frametext``` to display on the button, and then that value is used to update the framerate when settings are saved.

NotStanley4330: I do agree with it being a mess. I can look at just totally refactoring it today so it is so much more clear. It shouldn't break anything to just make it a predefined vector or enum. Best to just clean it up. Git blame tells me the code is 22 years old there so probably time to just fix it properly

marzojr: I'd say go ahead and refactor it as much as you want.

NotStanley4330: I'll refactor it this week and get it pushed back up!

NotStanley4330: Refactored!

PR #503, opened 2024-04-20
Background/ambient sfx music tracks are less annoying

### LINKED ISSUE #498: Background sfx music *tracks* are really annoying and the code has issues
In the Background_noise function https://github.com/exult/exult/blob/master/gamewin.cc#L158 there are some issues:

1. I really think the sfx tracks (4 (rainstorm), 5 (snowstorm), 6 (outside), 7 (nighttime)) are just noise with *some* little differences. IMO these should be disabled and only use the sfx for day and nighttime, defined later on.
2. We somehow overlooked that other music devices than digital, mt32(emu) or adlib can and should play the DangerNear und Dungeon tracks.
3. the wave sfx need their volume drastically reduced (nighttime AUDIO_MAX_VOLUME - 230, daytime AUDIO_MAX_VOLUME - 240)
4. we check twice for `if (hour < 6 || hour > 20)` (or `if (bghour < 6 || bghour > 20`) even though we have already made this the `bool nighttime`

IMO at least the nighttime and outside sfx music tracks should be disabled. And the code needs to make allowance for other music drivers to play DangerNear and Dungeon music.
(3 & 4 are easy enough to fix).

@wench @marzojr what are your thoughts?

DominusExult:
- bg tracks are disabled for the FMOpl driver, even though these were used in the original. They are way too noisy.
- DangerNear is now playing for all drivers
- nighttime bg track is disabled as it is too noisy
- bg sfx waves are playing on nighttime for all drivers
- bg sfx wave tracks are being played on SI as well

Fixes #498

### DIFF
--- a/gamewin.cc
+++ b/gamewin.cc
@@ -247,7 +247,6 @@ void Background_noise::handle_event(unsigned long curtime, uintptr udata) {
 	// Testing. When outside play birds during daytime or crickets at night
 	if (ava && !gwin->is_main_actor_inside() && currentstate != Dungeon) {
 		int                        sound;     // SFX #.
-		int                        volume;    // keep the volume down.
 		static const unsigned char bgnight[] = {61, 61, 255};
 		static const unsigned char bgday[]   = {82, 85, 85};
 		if (repeats > 0) {    // Repeating?
--- a/gamewin.cc
+++ b/gamewin.cc
@@ -221,9 +221,9 @@ void Background_noise::handle_event(unsigned long curtime, uintptr udata) {
 					}
 					laststate = RainStorm;
 				} else if (nighttime) {
-					// Disabled nighttime sfx track as it is very distracting
-					/*if (play_bg_tracks)
-						tracknum  = Audio::game_music(7);*/
+					if (play_bg_tracks) {
+						tracknum = Audio::game_music(7);
+					}
 					laststate = Nighttime;
 				} else {
 					if (play_bg_tracks) {
@@ -236,17 +236,18 @@ void Background_noise::handle_event(unsigned long curtime, uintptr udata) {
 		}
 	}
 
-	// Tests to see if day sfx track is playing, possible
-	// when the game has been restored
+	// Tests to see if Outside/Nighttime sfx tracks are playing,
+	// possible when the game has been restored
 	// and the Audio option was changed from OGG/MT32 to something else
-	// If nighttime sfx track is uncommented, also query whether track 7 plays
-	if (player && !play_bg_tracks && player->get_current_track() == 6) {
+	if (player && !play_bg_tracks
+		&& (player->get_current_track() == 6
+			|| player->get_current_track() == 7)) {
 		player->stop_music();
 	}
 	Main_actor* ava = gwin->get_main_actor();
 	// Testing. When outside play birds during daytime or crickets at night
 	if (ava && !gwin->is_main_actor_inside() && currentstate != Dungeon) {
-		int                        sound;     // SFX #.
+		int                        sound;    // SFX #.
 		static const unsigned char bgnight[] = {61, 61, 255};
 		static const unsigned char bgday[]   = {82, 85, 85};
 		if (repeats > 0) {    // Repeating?
--- a/gamewin.cc
+++ b/gamewin.cc
@@ -173,32 +173,20 @@ void Background_noise::handle_event(unsigned long curtime, uintptr udata) {
 	} else if (weather == 1) {
 		currentstate = SnowStorm;
 	} else if (nighttime) {
-		currentstate = Nighttime;    // Night time
+		currentstate = Nighttime;
 	} else {
 		currentstate = Outside;
 	}
 
 	MyMidiPlayer* player = Audio::get_ptr()->get_midi();
-	// Lets allow this for Digital Muisc and MT32Emu only,
-	// for MT32/FakeMT32 conversion as well.
-	// if (player) {
-	// if (player && player->get_ogg_enabled()){
-	if (player
-		&& (player->get_ogg_enabled() || player->is_mt32()
-			|| player->is_adlib())) {
+	// The sfx tracks only play for Digital Music, MT32emu, MT32/FakeMT32
+	// FMOpl is not sounding acceptable even though the original used it.
+	const bool play_bg_tracks
+			= player && (player->get_ogg_enabled() || player->is_mt32());
+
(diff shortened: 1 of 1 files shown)


gamewin.cc
```diff
@@ -218,66 +206,69 @@ void Background_noise::handle_event(unsigned long curtime, uintptr udata) {
...
-					tracknum  = Audio::game_music(4);
+					if (play_bg_tracks) {
+						tracknum = Audio::game_music(4);
+					}
 					laststate = RainStorm;
-				} else if (bghour < 6 || bghour > 20) {
-					tracknum  = Audio::game_music(7);
+				} else if (nighttime) {
+					// Disabled nighttime sfx track as it is very distracting
+					/*if (play_bg_tracks)
```
marzojr, line 225: I disagree with disabling this, at least like this. Maybe add a config option to disable it instead?

gamewin.cc
```diff
@@ -218,66 +206,69 @@ void Background_noise::handle_event(unsigned long curtime, uintptr udata) {
...
-			} else {
-				delay   = 4000 + rand() % 3000;
-				repeats = 0;
+	}
+
+	// Tests to see if day sfx track is playing, possible
+	// when the game has been restored
+	// and the Audio option was changed from OGG/MT32 to something else
+	// If nighttime sfx track is uncommented, also query whether track 7 plays
+	if (player && !play_bg_tracks && player->get_current_track() == 6) {
```
marzojr, line 243: Please add the check for track 7 anyway and remove/edit the comment. That way, if we later decide to uncomment the track 7 code (or if the config option I suggest is added), the code will work without needing further changes.

gamewin.cc
```diff
@@ -218,66 +206,69 @@ void Background_noise::handle_event(unsigned long curtime, uintptr udata) {
...
+					if (play_bg_tracks) {
+						tracknum = Audio::game_music(4);
+					}
 					laststate = RainStorm;
-				} else if (bghour < 6 || bghour > 20) {
-					tracknum  = Audio::game_music(7);
+				} else if (nighttime) {
+					// Disabled nighttime sfx track as it is very distracting
+					/*if (play_bg_tracks)
+						tracknum  = Audio::game_music(7);*/
```
marzojr, line 226: I disagree with disabling this, at least like this. Maybe add a config option to disable it instead?

marzojr (changes requested):

DominusExult:
Re-enabled the Nighttime track.
Before this is ready to be merged, @wench mentioned in #498 that his jukebox egg looping fix also fixes the bee cave outside of this.
```
// Don't override bee cave music with dungeon music.
			const bool notbees = !GAME_BG || curr_track != 54;
if (curr_track == -1 || gwin->is_bg_track(curr_track)
				|| (((currentstate == Dungeon && notbees)
					 || currentstate == DangerNear)
					&& !is_combat_music(curr_track))) {
```

And possibly this TODO, right?
```
// ++++ TODO: Need to come up with a way to replace repeating songs
// here, just so they don't loop forever.
```

wench:
> Re-enabled the Nighttime track. Before this is ready to be merged, @wench mentioned in #498 that his jukebox egg looping fix also fixes the bee cave outside of this.
> 
> ```
> // Don't override bee cave music with dungeon music.
> 			const bool notbees = !GAME_BG || curr_track != 54;
> if (curr_track == -1 || gwin->is_bg_track(curr_track)
> 				|| (((currentstate == Dungeon && notbees)
> 					 || currentstate == DangerNear)
> 					&& !is_combat_music(curr_track))) {
> ```
> 
> And possibly this TODO, right?
> 
> ```
> // ++++ TODO: Need to come up with a way to replace repeating songs
> // here, just so they don't loop forever.
> ```

Yep, should be fixed by my changes and I noticed that Skara Brae would have the same problem as most of the island is covered by eggs that all play the same music track. In fact i am going to guess the problem exists in lots of places in both games.

DominusExult:
> Yep, should be fixed by my changes and I noticed that Skara Brae would have the same problem as most of the island is covered by eggs that all play the same music track. In fact i am going to guess the problem exists in lots of places in both games. It wouldn't be reasonable to special case every location but my general looping fix is the right way to do things.

Thanks for fixing this!

So my pull request is ready once #508 is merged

marzojr (approved):

wench (approved):

Dragon-Baroque:
Sorry to come after the commit has been merged.

- I now have some warnings from GCC and CLang about `gamewin.cc` complaining that we may end up sending a random sound effect because neither the day ( at 254 ) nor the night branch ( at 251 ) are taken and that the `int sound;  // SFX#`  at 245 is not setting a value.
So I would like to replace 245 by `int sound = -1; // SFX #`. Not by 0 which is a valid SFX number.

- I would like also to protect the call to `play_sound_effect` and the computation of `repeats` and `delay` at 259..268 by a `if (sound >= 0)`, and place a reset of `repeats` to 0 into an `else` branch.

- I am a bit worried by the case `if (repeats > 0)` at 248. In effect if you are in a daytime _repeat_ the daytime music track check is bypassed and you replay the sound effect, _thus even if a music track started between the first emit of the sound effect and the current repeat_. But I have no good proposal for this.

Now about understanding better the purpose of `Background_noise::handle_event`.

- You expect to produce the  background _music_ tracks 4 RainStorm, 5 SnowStorm, 6 DayTime, 7 NightTime, but only on a faithful MT32 audio device, either using the digital OGGs or using the MT32 library, but not the FMOpl library. And only when no other regular - I mean not background - music tracks are playing.
 
- You expect to produce randomly the 61 Criquet sound effect at night, whether or not the any music track is playing.

- You expect to produce randomly the 82 and 85 Birds sound effects during the day, only when no music track is playing, and not when the audio device is a MT32 like. This is because the DayTime music track 6, already contains bird songs, is it ?

If this is so, I have the following comments to make :

- The test `if (player || play_bg_tracks)` at 187 is wrong, it should be `if (player && play_bg_tracks)`, otherwise the `play_bg_tracks` has no effect.

- The volume `AUDIO_MAX_VOLUME - 200` is very weak. I have tried various settings, using the headphones and the loudspeakers, and I think the setting should not go below `AUDIO_MAX_VOLUME - 32`.

Should I prepare a PR along those lines ?

Dragon-Baroque:
> I guess the volume is a very subjective thing :) maybe we can compromise on something in between? -50?

I changed to `AUDIO_MAX_VOLUME - 64`.

This is PR #511

Issue #491, opened 2024-04-09
Music Looping Behavior


DominusExult:
> Concerning the music looping.
> In the original, it went like this:
> As long as you are in the music area (like the stables at the start, a tavern or dungeon) the music would loop, even if you have gumps open or hold a conversation. Once you DO leave the area, the music would run out and then switch back to ambient sounds (unless another music is triggered).
> In Exult, there are two behaviors:
> -With looping enabled, the music loops endlessly, until you trigger a different music! The trigger that once you leave the music area there won't be more loops is missing. Thus you also never hear embient sounds (birds and stuff).
> -Without looping, the music just runs once and won't do so again unless you again enter the building in question or run over a very specific trigger.
> I compared it with the original. Right at the start in the stables, the song loops until you leave the stables, then it won't appear again. In Exult, the behaviors above happen.
> That's what I meant that the looping doesn't work like the original (in the original, the looping was quite clever).

> This is also incredibly annoying in SI with the Fellowship music never stopping, just hanging on a buzz.

from our old SF tracker https://sourceforge.net/p/exult/bugs/1990/

NotStanley4330:
I can poke at this one as well because I looked at some of the audio code recently. I definitely remember this one dissuading me from playthroughs in Exult vs DOSBox previously, as I got really sick of the over blasting music that made it seem like Ultima 6 all over again lol. I'd bet we can do a similar thing to weather effects where you track distance from an egg and just have it no longer loop once you get far enough away.

DominusExult:
relevant is this as well https://sourceforge.net/p/exult/patches/104/

And some other music track bugs https://sourceforge.net/p/exult/bugs/1839/ https://sourceforge.net/p/exult/bugs/1838/ and https://sourceforge.net/p/exult/bugs/1843/ - Might need to open a separate issue for those.

wench:
we interpret the music egg's looping flag to mean loop endlessly but it would make sense if it actually meant loop while the egg is in proximity. As it is there is no way to stop a track from looping once it has started so I'll need to make changes to the music player. Asigning this to myself

DominusExult:
D‘oh, I never realized that THIS is what is odd about the looping music, of course it should only loop around the egg.

wench:
I should have this ready to push in a day or so. I need to do some more testing to make sure it is working as intended (at the moment I'm not convinced it is)

wench:
[...]

PR #508, opened 2024-04-28
New music looping behaviour

wench:
This pull request is for changes related to #491  altering music looping so music will not continuously loop when too far from Jukebox Eggs that started looping music


DominusExult:
For documentation's sake can you elaborate on what limited and auto does?
Limited uses your change I guess, stopping the loop when the jukebox egg is too far away, right?
And auto?

wench:
> For documentation's sake can you elaborate on what limited and auto does? Limited uses your change I guess, stopping the loop when the jukebox egg is too far away, right? And auto?

Auto is the intended new be and will be the new default, 
Music will continue to loop while in the active area of the egg and when leaving the active area the repeat flag on the song is cleared so the song will will continue to play but will not repeat when it reaches the end

Limited will also loop the song while in the active area but when leaving the active area the song will immediately stop.

Config files will be interpreted with old "no" setting becomes "never", and any other old value becomes "auto"

wench:
I am not fixing the CodeFactor issue with Egg_object::test_unhatch() as that method is mostly a copy Paste of Egg_object::is_active() with inverted logic. It seems better if the new method stays similar to the old one and I doubt anyone cares about complex methods as the old method has a worse complexity rating

marzojr:
> I am not fixing the CodeFactor issue with Egg_object::test_unhatch() as that method is mostly a copy Paste of Egg_object::is_active() with inverted logic. It seems better if the new method stays similar to the old one and I doubt anyone cares about complex methods as the old method has a worse complexity rating

To be fair, I disagree a lot about how these tools determine a method's complexity: if you have a function consisting of a switch with enough cases that dispatches each case to a different function will result in a "complex code" warning, despite the fact that any human looking at it will trivially understand what the code does.

I would suggest ignoring such warnings unless the code really is difficult to follow.

wench:
One thing I forgot to put in commit message is in DEBUG build with Egg displayed turned on The Active area of Jukebox_eggs will be drawn in the world as well as the Avatar's footprint, This tends to draw lines all over the screen but it is useful for testing what the eggs are doing.

audio/Audio.cc
```diff
@@ -215,7 +215,7 @@ Audio* Audio::get_ptr() {
 Audio::Audio() {
 	assert(self == nullptr);
 
-	string s;
+	string s, newval;
```
marzojr, line 218: Why not declare newval where it is first used (line 234)?

wench: I put the declaration in first before using it

objs/egg.cc
```diff
@@ -1111,6 +1386,12 @@ void Egg_object::hatch(
 
 	/* end hack */
 
+	// Must not try to hatch an unhatchble egg unless unhatch() was called and
+	// removedthe hatched flag
```
marzojr, line 1390: `removedthe` -> `removed the`

objs/egg.cc
```diff
@@ -1121,7 +1402,8 @@ void Egg_object::hatch(
 			// We have been deleted, so just leave.
 			return;
 		}
-		if (flags & (1 << static_cast<int>(once))) {
+		// unhatchble eggs must not be removed here
```
marzojr, line 1405: unhatchable

objs/egg.cc
```diff
@@ -858,6 +1030,109 @@ bool Egg_object::is_active(
 	}
 }
 
+// test_unhatch pretty much inverts all the criteria compared to is_active
+// thiswas added for improve jukebox_egg behaviour to give the eggs a proximity
+// near and far trigger unless other egg types are modified this code should not
+// effect anything except jukebox_eggs
+
+bool Egg_object::test_unhatch(
```
marzojr, line 1038: I wonder if the logic we have in weather effects for weather eggs should not use this instead?

marzojr (approved):

Issue #514, opened 2024-05-05
Clang-format output depends also on CLang Version


Dragon-Baroque:
I mean, it depends first on the `.clang-format` at the top of the Git tree, but it depends _also_ on the version of CLang.

While using Linux Fedora 39 and CLang 17.0.6, I applied `clang-format` globally to my commits and I felt like I only reformatted the updated lines of my updated files. Not so when I switched to Fedora 40 and CLang 18.1.1, a global `clang-format` modified a lot more files than just the updated files and a lot more lines than the updated lines in the updated files. Fortunately I could install a non default CLang 17 whose `clang-format` went back to the reformats I was used to.

On the other hand, wench has committed twice on April 27 using a `clang-format` which looks like coming from CLang 18.
@wench, could you perform a `clang-format --version` ? I get this :
```
clang-format --version
clang-format version 18.1.1 (Fedora 18.1.1-1.fc40)
clang-format-17 --version
clang-format version 17.0.6 (Fedora 17.0.6-7.fc40)
```

- If I apply a CLang 18 `clang-format` to `master` as of today, I get 11 files being reformatted, mostly line cuts and realigns but a few line joins too. Almost all look reasonable.
- If I apply a CLang 17 `clang-format` to `master` as of today, I get 35 files being reformatted instead, some being a revert of wench commits of April 27. For example compare `usecode/intrinsics.cc` table `offsets`, lines 2015+, from `master` of today ( `clang-format` of CLang 18, I suppose ) with `master` of April 23 ( `clang-format` of CLang 17 ).

I am afraid that if we developers apply blindly the ( currently installed on our build site ) `clang-format` to our commits, without following a clear policy that the CLang should be of a _clearly stated_ version, we will end up polluting `master` with massive back and forth reformat changes. @DominusExult, @marzojr, could you please define such a policy ?

DominusExult:
Wow, good points. I didn't even think of this.
Without installing it through a 3rd party package manager since Apple doesn't package it with its clang.
So far I used https://clang-format-configurator.site which uses version 18.1.3 as I now see after you made the point about versions.
I have no idea how to proceed.

marzojr:
Since the latest official release is [clang 18.1.5](https://releases.llvm.org/), and the last 17 version release 17.0.6 from November last year, I would say that we should not run clang-format if it is less than version 18.

Maybe we should add something like [ImgBot](https://imgbot.net/) that formats on push and creates a pull request? And maybe do a format check when a pull request is made as well?

[...]

PR #515, opened 2024-05-06
Switch Exult to clang-format version 18

Dragon-Baroque:
This completes the replacement of `clang-format` version 17, used up to mid-April 2024, by `clang-format` version 18, by agreement in Issue #514.

**For reviewers :** The one change that I feel funny is `new MT32Timbre* [128] {};` packed into `new MT32Timbre*[128]{};`.

I can obtain  `new MT32Timbre*[128] {};` instead with `SpaceBeforeCpp11BracedList: true` instead of `false` in `.clang-format`.

- But then all brace initializers get a new space like `const std::array frames{` becoming `const std::array frames {` or `actions = new Actor_action*[5]{` becoming `actions = new Actor_action*[5] {` in `actions.cc`. In total this makes more that 500 changes in more than 60 files.
- Also in `actions.cc` there are initializers with parentheses and initializers with braces. With this spacing the lookout becomes inconsistent, I get `: frames {f}, index(0), speed(spd), obj(weak_from_obj(o)) {` where the braced initializer gets a space but the parenthesized initializers do not.

### DIFF
--- a/audio/Midi.cc
+++ b/audio/Midi.cc
@@ -157,7 +157,8 @@ void MyMidiPlayer::start_music(int num, bool repeat, std::string flex) {
 		// No midi driver or bg track and we can't play it properly so don't
 		// fall through
 		if (!midi_driver
-			|| (!is_mt32() && Game_window::get_instance()->is_background_track(num)
+			|| (!is_mt32()
+				&& Game_window::get_instance()->is_background_track(num)
 				&& flex == MAINMUS)) {
 			return;
 		}
--- a/audio/midi_drivers/ALSAMidiDriver.cpp
+++ b/audio/midi_drivers/ALSAMidiDriver.cpp
@@ -160,8 +160,7 @@ int ALSAMidiDriver::open() {
 			pout << clt_id << ":" << prt_id << " : [ " << clt_name << " : "
 				 << prt_name
 				 << (prt_write_use > 0 ? " ], in use, " : " ], available, ")
-				 << prt_midi_channels << " channels"
-				 << ", RW capability"
+				 << prt_midi_channels << " channels" << ", RW capability"
 				 << ((prt_capability & SND_SEQ_PORT_CAP_READ) != 0u ? " R" : "")
 				 << ((prt_capability & SND_SEQ_PORT_CAP_WRITE) != 0u ? " W"
 																	 : "")
@@ -198,9 +197,8 @@ int ALSAMidiDriver::open() {
 			pout << std::endl;
 			return 0;
 		}
-		perr << "ALSAMidiDriver: "
-			 << "Can't subscribe to default MIDI port [" << seq_client << ":"
-			 << seq_port << "]";
+		perr << "ALSAMidiDriver: " << "Can't subscribe to default MIDI port ["
+			 << seq_client << ":" << seq_port << "]";
 		if (identify_port(seq_client, seq_port)) {
 			perr << " [ " << clt_name << " : " << prt_name << " ]";
 		}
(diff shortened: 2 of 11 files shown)


DominusExult:
I also think that the handling of parentheses is odd at times and doesn't appeal to my sense of order...
(not the first time this is changed thus, I just noticed this the first time now)
```
} else if (
					!play_bg_tracks
					&& (!player || player->get_current_track() == -1)) {
```

```
bool is_background_track(
			int num) const;        // ripped out of Background_noise
```

Dragon-Baroque:
> I also think that the handling of parentheses is odd at times and doesn't appeal to my sense of order...

Me neither, though I should point out :

- That the console output ( in Linux, when I use git diff ) and the diffs in GitHub expand the tab **by default** to 8 spaces whereas any source editor and `clang-format` expand the tab to 4 spaces. This makes the sources appear wrongly to expand very quickly to the right. I have not found any way to convince the console to convert the tab to 4 spaces. I have however altered the **tab setting of GitHub** to 4 spaces ( Profile -> Settings -> Appearance -> Tab Size preference ).
- That I tried to join the two first lines in `gamewin.cc` of your sample to this ( I have corrected in these copy-pastes the 4 to 8 spaces discrepancy )
```
            } else if (!play_bg_tracks
                    && (!player || player->get_current_track() == -1)) {
```
but when I applied `clang-format` over it, it was reestablished as ( as it is now ) :
```
            } else if (
                    !play_bg_tracks
                    && (!player || player->get_current_track() == -1)) {
```
where the various subcomponents of a pair of parentheses are consistently right indented of 8 spaces ( 2 tabs ).

wench (approved):

DominusExult:
> * I have however altered the **tab setting of GitHub** to 4 spaces ( Profile -> Settings -> Appearance -> Tab Size preference ).

That's very helpful! :)

Not happy with the parentheses and I think clang format does it wrongly in that case.

DominusExult:
if clang-format is changed from `AlignAfterOpenBracket: AlwaysBreak` to `AlignAfterOpenBracket: BlockIndent` it looks better IMO.
```
} else if (!play_bg_tracks
           && (!player || player->get_current_track() == -1)) {
```
(this is just an approximation as again the Github tabs things shows it differently in the markdown than it is shown in the file)

Dragon-Baroque:
> if clang-format is changed from `AlignAfterOpenBracket: AlwaysBreak` to `AlignAfterOpenBracket: BlockIndent` it looks better IMO.

Agreed. Problem however is, if I apply this to Exult in full, I get 385 files modified for more that 5000 changes.
Essentially all function calls, functions and method declarations _that take more than one line_ get their closing parenthesis on a new line. Like this ( from `usecode/ucsymtbl.h` ) :
```
    Usecode_scope_symbol(
            const char* nm = "_usecode_", Symbol_kind k = table_scope,
            int v = -1)
            : Usecode_symbol(nm, k, v) {}
```
becoming
```
    Usecode_scope_symbol(
            const char* nm = "_usecode_", Symbol_kind k = table_scope,
            int v = -1
    )
            : Usecode_symbol(nm, k, v) {}
```
This, even if I have to admit that this looks nicer, I would not apply such a change unless we all agree for it.

DominusExult:
yeah, that is something to collectively decide IMO

Dragon-Baroque:
If big effect `clang-format` changes are to be agreed, I would like to merge the present commit as it is.

So that I can migrate my own Git branches to `clang-format` 18.

DominusExult:
Oh yes, go ahead, I'll bring it up in the issue about clang-format versions

Issue #519, opened 2024-05-10
Allow dragging of modal gumps


wench:
Exult replicates original behaviour and does not support dragging modal gumps to move them. Geneally this is fine but it can cause trouble if the size of the game area is less than 320x200. This was never a concern for the original as it always ran at 320x200 but exult can run at other resolutions. In particular the apply button on the video options gump is right at the bottom so it can go off screen if the game area is less than 200 pixels high so if for some reason the game area is too small it may be impossible to change settings to something better from within exult.
I got thinking about this so people could run exult at 1080p (actually 1600x1080) and have perfect aspect correct scaling of a 320x180 game area with 6x vertical and 5x horizontal scaling factors. Losing those 20 vertical pixels shouldn't cause problems in game (it ends up being 10 pixels at top and bottom so a fraction over a tile) but might for modal gumps. For 320x200 perfect aspect correct scaling a 1200p display is needed and while I have 2 of them (16:10 workstation displays) they are not particularly common so I wantto make this a useful option for all.

If anything adding UI consistency allowing all gump types to be draggable would be nice
This would also allow us to have oversized gumps bigger than 320x200
I would add a config option to disable this behaviour for purists.
Even if disabled by config I would probably always allow it for gumps that clip offscreen
An alternate idea would be to add edge scrolling for offscreen gumps, but as this is not a standard UI interaction in u7 it seems like a bad idea. (crazy idea add edge scrolling to the game world too)
I am currently doing a lot of things to gumps while working on #502 so I will probably in this in the commits for that.

Issue #502, opened 2024-04-14
Audio Volume Mixer WIP


DominusExult:
I've created a new gump to control the volume of music, sfx and speech.

So far it can only read and write the settings of exult.cfg.

Atm I'm stumped to actually move the slider via the scroll buttons, dragging with the mouse or mousewheel. This is out of my meager programming knowledge and I'm hitting a wall.

Furthermore this is not hooked up to the actual volume of these sources.
SFX can probably easily be done via `AUDIO_MAX_VOLUME`, music maybe via `volume_curve`. AFAICT there is nothing to control the volume of voc files.

I have this branched in my fork of Exult at https://github.com/DominusExult/exult/tree/volume_mixer.
Pressing Shift+V will open the gump.

![image](https://github.com/exult/exult/assets/3313896/fb34c13d-4136-4a7e-ab74-19bbeafb9678)

NotStanley4330:
I pulled this down and will look at getting the sliders working. I can also make it so it controls master SFX volume as I know how that works fairly well.

wench:
I'll add in code that will allow proper realtime volume control of music . volume curve isn't really appropriate for this. Not sure if ogg music and midi should have separate controls. Midi could have separate controls per device but that doesn't really seem necessary

wench:
Thinking the gump should only control music volume for the current music device or oggs but the config file should probably store things separately per device to allow people to tune volume per device by switching between them.

DominusExult:
> Thinking the gump should only control music volume for the current music device or oggs but the config file should probably store things separately per device to allow people to tune volume per device by switching between them.

This might be a bit over the top for the normal user but probably welcomed by more "serious" users.
So I guess `<volume>` should be its own section under `<audio> `with several `<volume_DRIVERNAME>` and `<volume_sfx>`, `<volume_speech>`

wench:
Sounds good to me. As I add music volume support to the Audio code I'll update the Audio code so it will use those config settings to set music volume upon startup. I'll fork your branch and make a pull request on your repository when done

DominusExult:
I've updated my fork with main Exult and added a button to the Audio Options gump.
(name of the button and the used shortcut key are open for discussion :))

The Audio Options gump could use just a tiny bit more space or I need to minimize the gap between the section by just 1 pixel to have a wider gap between the mixer button and the audio section.

![image](https://github.com/exult/exult/assets/3313896/70192437-bfe6-40a1-9df9-677f92b68fcc)

[...]

PR #529, opened 2024-07-02
Volume mixer and Gump Changes

wench:
This is PR to pull all changes from the volume_mixer branch into master for #502 .

Originally it was just meant for the new volume mixer gump but as I got into it and was knee deep in Gump code I made a number of other changes to gumps in general.

One of the more important changes to Gumps was to add a new base class shared by Gumps and Widget to provide an interface that will allow Widgets to be parents of other widgets. In particular this was to allow complex widgets to contain instances of Gump_button like the new Slider_Widget class.

At the same time I added the Input handling Interface used by Modal_gumps into the new base class so It is easy to pass Input recieved by a Modal Gump to widgets without any special processsing first. As part of this process I removed the usage of integer magic numbers to refer to Mouse Buttons. It was very unclear what code like `button == 1` actually meant (is it left, right or something else, it is actually the Value that SDL uses for Left). I updated  everything that used the Modal Gump input interface to use an enum for Mouse buttons instead. Gump Manager now converts mouse buttons numbers from SDL into the Needed Enum values before sending the input to the current Modal Gump. This change has no effect on Non modal gumps as non modal gump do not handle their own input, their input is handled by the main game Input handler and i didn't touch that. 

Also while messing with Modal Gump Input I implemented the ability to drag Modal Gumps as mentioned in #519 This new feature that is a departure from Original behaviour and is controlled by Config setting `config/gameplay/modal_gump_dragging` This setting can take 3 values
- Never = Dragging of Modal gumps is not allowed as in original game behaviour
- Always = Dragging of Modal gumps is always allowed like regular non modal gumps
- Offscreen = This is the default and only allows Modal gumps to be dragged if they are partially offscreen (the intent of the change for incase Exult is run with Video settings that cause the Game screen to be much smaller than 320x200, the video settings gump will still be usable)

It should be noted this only applies for Gumps that pass Unhandled Mouse input down to the Modal Gump class. Certain gumps like Text Gumps, Book Gump and Sign Gump do not inherit from Modal Gump and are unaffected by this change and cannot be dragged. ItemMenu_gump is a Modal Gump but as it handles all input it receives it also cannot be dragged at this stage.  Individual Gump classses can override `virtual bool is_draggable()` or set `Modal_gump::no_dragging=true` to prevent dragging that gump

As part of implementing the Volume Mixer I created a new Slider Widget class. The Slider Widget was created by taking all the important functionality of the Slider Gump and turning it into a self contained Widget. I updated Slider Gump to use an instance of Slider Widget for its Functionality. As well as using it multiple times in the New Mixer Gump class. New Slider Gump behaviour is slightly different to old as Some behaviour seemed to me like they were slightly bugged so I fixed them in the slider widget code. If this is a problem all the old slider gump code is #ifdefed out and easy to restore if needed. New Slider widget has code to support a Logarithmic slider but it isn't used unless specified in the constructor arguments, chances are there is no need for this but it was just an idea i had and it was easy to implement so I did it

As part of the Volume Mixer Gump I added the ability for Modal Gumps to have a basic Procedurally generated background removing the need for shapes to be created for new Modal Gumps that are just flat brown boxes with bevelled 3d edge. I updated all the Option Gumps to use this system and removed their Shapes from data. They look visually identical to the older versions with a shape background. The intent behind this was to allow dynamic resizing of gumps to match their content. The mixer gump can show 3 or 4 sliders depending on Audio settings and resizes to fit. As part of this change I added the ability to recolour the Checkmark background. By default it is the same green that all our gumps used But I added code to allow changing the checkmark background to use any of the other colour ramps in the palette if desired instead of its default green. The code to do this can be applied to any shape using the ShapId::set_palette_transform() method I added a little while ago. It can be used to selectively recolour only part of a shape. For instance I have been playing around with trying to recolour the Colour of the Cloak On The Avatar's Shape and results are quite nice, but that is something for a new issue or a discussion.

Modal Gumps can also now display a popup message above the gump (call `Modal_gump::SetPopupMessage()`). This for now is only used by The Mixer Gump but it might be useful for other gumps to show non-fatal error messages and other messages that do not need user acknowledgement. Only works if the gump calls `Modal_gump::paint()`, many gumps did not they directly called `Gump::paint()` but I fixed them so they now call paint on their direct parent.

The Volume Mixer  required changes to the Audio system so All Audio Types can Have their Volumes controlled and changed. config file settings for volumes (0-100, default 100) are
- `config/audio/effects/sfx_volume` for sfx volume
- `config/audio/speech/speech_volume` for speech
- `config/audio/midi/volume_ogg`  for ogg music
- `config/audio/midi/volume_[drivername]` for midi drivers (allows changing each midi driver independently but usually there will only be a value for the selected midi driver)
 
Exult will create these values in exult.cfg at startup

I also went through and ensured that each Audio type is properly classified when played. Mostly everything was already ok BUT The Thunder in the SI intro was classified as Speech but I changed it to now be classified as SFX. I'm pretty sure this was just an old hangover from when I first implemented the intro and we didn't even have digital SFX support yet, only speech. But it did match original game behaviour as the original games just did not support non midi sfx so the thunder was treated as speech.

I added ability for some small amount of error handling when playing music. Methods now return true on success when playing music and it is possible to retrieve the filename of an ogg that failed to play. If oggs are enabled it is now possible to indicate that you want to play only the ogg or midi version of the track with no fallback on failure.

For discussion about the Volume Mixer Gump itself see issue #502


wench:
Ignore this, I messed it all up making a new PR

PR #534, opened 2024-07-12
Using string_view on config/xml/utils

marzojr:
Not fully using everywhere, but enough to make a good start.

### DIFF
--- a/.clangd
+++ b/.clangd
@@ -33,7 +33,7 @@ If:
   PathMatch: .*\.h
   PathExclude: [ios/.*, mapedit/tools/mockup/.*, mapedit/tools/smooth/.*]
 CompileFlags:
-  Add: [-xc++-user-header, -std=c++17]
+  Add: [-xc++-user-header, -std=c++17, -DVERSION="1.9.0git"]
 ---
 If:
   PathMatch: [.*\.c]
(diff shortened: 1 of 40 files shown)


wench:
We already have some case insensitive string comparison functions for null terminated strings in istring.h/cc, you might want to put iequals and safe_tolower  in there

marzojr:
So I have been working on this a bit. I have one absolutely giant commit which I am not sure I can make smaller because it is a mess of interconnected changes.

And I am thinking that maybe I want to work on adding cmake support first, with the intent of using [fmtlib](https://github.com/fmtlib/fmt), [scnlib](https://github.com/eliaskosunen/scnlib), and maybe [spdlog](https://github.com/gabime/spdlog/). They would all help improve some of the things I am working on on both this pull request, and on #477.

Another issue is the overlap of this and #477. Both make several similar changes on similar lines, and merging these is going to be a pain.

So I am thinking of creating a dev branch that (1) adds cmake support, (2) starts using fmtlib and scnlib in places where snprintf/scanf/etc are used, and (3) merge this pull and #477 onto said branch to work.

That said, I think I will add a commit to this that changes the way doubles are read from/written to config because the current method has issues with localization.

DominusExult:
That's a sound way to move forward with this IMO

PR #537, opened 2024-07-17
Adding Mouse and Touch support to Cheat screen and other cheat screen changes

wench:
The PR adds mouse and touch support to the Cheat screen. Primarily it does this by Creating hot spots for each menu item that can be clicked. The hot spots are the Bracketed Key indicators on each menu item. For small touch devices the hot spots may be a bit small so the pre-existing method of using the On screen keyboard to select menu item can still be used. As part of adding hot spots I added highlights to the hotspots for when the mouse hovers over them. 

I also added an additional highlight that will highlight the entire menu item when it is selected. This only works for Menu items that do not switch to a new menu. Hover highlight is red, activation highlight is blue and both at the same time give purple

Double clicking/tapping will now act as pressing the return key provided the clicks and taps were not on a hotspot

An addition to a recent change I made to use arrow keys in the Cheat Screen, I added the ability to increment and decrement numeric value inputs using right (inc) and left (dec). There being no input is treated as zero when incrementing or decrementing. There are limits imposed when incrementing and decrementing based on the current mode, 

For the mouse I made it so the mouse wheel will simulate left and right key presses. This makes it possible to use the cheat screen entirely with the mouse except for text input like change name. It's not very efficient to select numbers with the mouse wheel but it is at least now possible to use it without using the keyboard for most functions (this is a minor accessibility feature for people like myself with disabilities preventing normal keyboard+mouse use).

Related to that, when using Touch, Swipes across the screen will also act as arrow key presses (up, down, left, right). Longer swipes will cause repeated simulated key presses. It is setup to give 1 keypress for every 7.5% of the screen a swipe goes across. You should get about 13 key presses if swiping across the entire screen, 

I made some menu layout changes so the [ESC] exit options are always in the same place for every menu. This allows for quick exiting for deep in the menus without moving the mouse. Just hover over [ESC] and left click till you are out of the cheat screen

Previously Tapping anywhere on the cheatscreen would toggle the OS onscreen touch keyboard. I have changed things slightly. Tapping on hotspots will not toggle the OSK. The keyboard will also not be shown if a tap occurs in a 20 pixel border around the edge of the screen so swipes from the edge of the screen will not trigger the keyboard. If the OSK is shown when doing a swipe, it will get hidden.

In The Version info shown by the cheat screen's initial menu, I added the compile date and time as an additional line of info. I mostly added this so I could know for sure that my phone was actually using the latest Android build I compiled. I had no end of troubles trying to debug Exult on Android and this was just a simple sanity check.

I also Changed VersionGetGitRevision to return string_view and made the necessary change to the Cheat screen so this would work.

I did some general clean up of some parts of the cheat screen code. There is more I could do, but I don't think re-engineering the entire thing is really worth it,

### DIFF
--- a/cheat_screen.cc
+++ b/cheat_screen.cc
@@ -20,6 +20,11 @@
 #	include <config.h>
 #endif
 
+// Disable the gcc warning because we cannot fix it in SDL's headers
+#if defined(__GNUC__)
+#	pragma GCC diagnostic push
+#	pragma GCC diagnostic ignored "-Wold-style-cast"
+#endif
 #include "cheat_screen.h"
 
 #include "Configuration.h"
@@ -49,14 +54,9 @@
 
 // #define TEST_MOBILE 1
 
-// Disable the gcc warnig because we cannot fit it the problem is in the Macro
-// from SDL
-#if defined(__GNUC__)
-#	pragma GCC diagnostic push
-#	pragma GCC diagnostic ignored "-Wold-style-cast"
-#endif
 static const Uint32 EXSDL_TOUCH_MOUSEID = SDL_TOUCH_MOUSEID;
 
+// renable the warning that was disabled above
 #if defined(__GNUC__)
 #	pragma GCC diagnostic pop
 #endif
@@ -218,6 +218,7 @@ void CheatScreen::show_screen() {
 	Mouse::Mouse_shapes saveshape = Mouse::mouse->get_shape();
 	Mouse::mouse->set_shape(Mouse::hand);
 
+	buttons_down.clear();
 	// Start the loop
 	NormalLoop();
 
@@ -620,12 +621,23 @@ bool CheatScreen::SharedInput() {
 					break;
 				case SDL_FINGERDOWN: {
 					CERR("SDL_FINGERDOWN");
+					buttons_down.insert(button_down_finger);
 					if ((!Mouse::use_touch_input)
 						&& (event.tfinger.fingerId != 0)) {
 						Mouse::use_touch_input = true;
 					}
 					break;
 				}
+				case SDL_FINGERUP: {
+					CERR("SDL_FINGERUP");
+					// Get iterator for first instance of button_down_finger and
+					// erase it from the collection
+					auto it = buttons_down.find(button_down_finger);
+					if (it != buttons_down.end()) {
+						buttons_down.erase(it);
+					}
+
+				} break;
 					// Finger swiping converts to cursor keys
 				case SDL_FINGERMOTION: {
 					gwin->get_win()->screen_to_game(
@@ -686,7 +698,7 @@ bool CheatScreen::SharedInput() {
 							event.button.x, event.button.y,
 							gwin->get_fastmouse(), gx, gy);
 					CERR("SDL_MOUSEBUTTONDOWN( " << gx << " , " << gy << " )");
-
+					buttons_down.insert(event.button.button);
 					if (event.button.button == 1) {
 						simulate_key = CheckHotspots(gx, gy);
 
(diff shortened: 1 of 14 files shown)


DominusExult:
Gonna take a look tomorrow. It really helps that you got it running on your Android, so you notice the need to place everything in the upper half/third of the screen ;)

wench:
Unless I messed something up, it seems like the business schedule menu was an exception, the menu is drawn at the bottom and the info at the top. I changed the way it is drawn but I don't think I changed the layout, but now with  hotspots it is more usable, you don't have the OSK blocking everything.

DominusExult:
Just tested on macOS and stumbled over that clicking on `Set Time` kind of made everything unresponsive in the cheat screen (neither click nor keyboard entry works anymore).
IOS I still need to test

DominusExult:
tested on iOS and it just works :)

However if you press [ESC] Exit in the cheat screen to exit the screen, it will register as a "Pathfind with long touch" (enable in Game Input gump) event. This happens on both iOS and macOS

DominusExult:
Sometime ago I fixed something similar on mobile: 
If you closed the notebook (checkmark top left on screen) it would „bleed through“ and open whatever shortcut bar item was beneath the checkmark.
Need to find what I did then, though

DominusExult:
touch_pathfind is in https://github.com/exult/exult/blob/master/exult.cc#L1693 (and https://github.com/exult/exult/blob/master/gamewin.cc#L1891 for the effect). I guess you would need some check in exult.cc, similar I did for shortcutbar gump and checkmarks in 79e4fd15b

wench:
Ok, that makes some sense. Button Down vs Button Up handling. Cheat Screen does not consume any mouse button up events

wench:
OK fixed the Long press issue. Cheat Screen will now wait for and held keys, buttons and fingers to be released before returning to the game. It will wait silently for 1 second for the up events then it will print a message detailing what it is waiting for and will allow pressing escape to immediately exit and return to the game with the unhanded up events.

I suspect this same issue will present itself in other places due to it not being needed to do anything with the up events generated after handling a down event. 

I also added a change so when the cheat screen shows the ask to quit gump if the user tries to close Exult's Window, the Gump will no longer be shown over the game world but will be shown over the cheat screen

DominusExult:
Tested again and this way of using the cheat screen is very enjoyable (especially on iOS it adds a lot of usability value).

Unrelated to this I found that the business activity screen needs more work on mobile (NPC -> (>0) -> Buisness activity -> A -> what should be in the upper half is in the lower half)). This is most likely my fault from some time ago and will fix once your changes are merged.
image

DominusExult:
And a left over from your changes to `+/-` and `*` in the cheat screen. When you can't use arrow up (first screen) or arrow down (always), using these will enter R or Q respectively.

wench:
Fixed the issue with the arrow keys not being shown as arrow but as other unrelated characters.

That set activity display is worse that I thought (i never checked that feature. I added `// #define TEST_MOBILE 1` to the top of cheat_screen.cc uncomment it to force mobile display on desktop systems to make it easier to fix things like the activity selection without needing to actually use a mobile device

Dragon-Baroque:
You left a minor spelling mistake using `SDL_KeyCode` instead of `SDL_Keycode` in `cheat_screen.h + cc`.
I noticed that while rebasing my Exult SDL 3 branch onto the new master - with this development.
Nice touch, thank you, the highlight of the labels when the pointer hovers over them.

Do you allow me to fix the mistake ?

wench:
Fix away. Do whatever you like

Dragon-Baroque:
Done by commit 00880da31.

And thanks also for you allowing me to lift your Drag and Drop code for Windows : It is now being pushed as the branch `sdl3-droptext-windows` of my fork of SDL 3 https://github.com/Dragon-Baroque/SDL/tree/sdl3-droptext-windows.
I hope to Pull Request it to SDL 3 in the coming days.

PR #564, opened 2024-09-20
Add AppStream metadata and release info

fpiesche:
These files are what the FreeDesktop project describes as "catalog metadata", for display in graphical package managers. Primarily this would be used for the Flatpak distribution to render the project's page on the FlatHub listing (e.g. https://flathub.org/apps/io.itch.amcsquad.amcsquad), but if a Linux distribution vendor opted to create a system package for Exult they could also include these files so that package managers can display the same metadata.

The release info here is currently manually maintained and again would be used to display release information on Flathub; I'm currently working on a GitHub Action to automatically update the release info on AppStream catalog metadata files like these when a release is created. I'll add another PR to implement this when I'm done with that particular bit of yak shaving... :)

### DIFF
--- a/desktop/info.exult.exult.metainfo.xml
+++ b/desktop/info.exult.exult.metainfo.xml
@@ -9,7 +9,7 @@
   <summary>A cross-platform engine for the Ultima 7 games</summary>
   <metadata_license>CC0-1.0</metadata_license>
   <launchable type="desktop-id">info.exult.exult</launchable>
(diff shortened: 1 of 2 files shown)


DominusExult:
Can you exchange Ultima 7 with Ultima VII? I need to change this in various other occurrences (like Readme.md where this was copied from).
I've fixed that elsewhere a while ago but it likes to creep back in :)

DominusExult:
Thank you for this and the updated icons!

Issue #572, opened 2024-10-01
SI: Topo's schedule has "Eat at Inn" activity with wrong location


mbanck:
Topo (the apprentice of the artisan in Moonshade, who buys gems off you) has this schedule:

```
Time          | Activity            | Location
--------------+---------------------+-----------
12 AM -  3 AM | Sleep               | 2358,1736
 3 AM -  6 AM | -                   | -
 6 AM -  9 AM | Wander              | 2377,1926
 9 AM - 12 PM | -                   | -
12 PM -  3 PM | Eat at Inn          | 2359, 1908
 3 PM -  6 PM | Wander              | 2376,1928
 6 PM -  9 PM | -                   | -
 9 PM - 12 AM | -                   | -
```

The problem is the location for the "Eat at Inn" activity is in the upper left corner (large wine barrel) of the artisan's workshop. In the original, he is standing there and walking around a bit, but if ask him about buying something or selling gems, he says he´d be glad to do that when he is at work.

In exult, he just stands there at those coordinates awkwardly partly on the wine barrel, see attached screenshot. I think the location for the "Eat at Inn" activity is wrong and should be changed to the Blue Boar Inn
![Screenshot from 2024-09-30 21-36-12](https://github.com/user-attachments/assets/e1dd50ce-5e14-4a03-a8d4-b14b4c735919)

PR #573, opened 2024-10-01
Sifixes: Add fix for Topo's schedule.

mbanck:
Topo has schedule 26 (eat at inn) for 12PM-3PM, but the location of that business activity is where he works (Ducio's shop), so if you talk to him and ask him about buying something or selling him gems he will claim that he is not at work - which is technically correct cause he is eating at the inn... Fix that by resetting the location of his eat at inn schedule to the Blue Boar.

Fixes: #572

### DIFF
--- a/content/sifixes/src/npcs/topo.uc
+++ b/content/sifixes/src/npcs/topo.uc
@@ -1,0 +1,31 @@
+/*
+ *
+ *  Copyright (C) 2024  The Exult Team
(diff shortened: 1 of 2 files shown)


DominusExult:
Thank you! sorry for taking so long but I lost sight of it :(

mbanck:
Thanks for looking at it!

PR #616, opened 2025-01-19
Fix version number and missing XML closing tag

fpiesche:


wench:
No I will not merge these into the 1.10.x branch as that branch is not undergoing active development right now and we have no plans to make another 1.10 release. The 1.10.x branch should remain static unless a new bugfix update needs to be done
I wont close this in case someone else wants to merge this 
If these changes need to be applied to master then make another PR for that

fpiesche:
Ah, fair enough! I was a bit confused by the lack of the v1.10 release info on the master branch; will rebase to that and open a PR over there.

PR #617, opened 2025-01-21
Update Linux desktop data

fpiesche:
Two major changes:

- Merge the separate `releases.xml` and the main metadata file. While this is part of the FreeDesktop specification, Flatpak does not currently [properly support](https://github.com/flatpak/flatpak/issues/6057) external release metadata, so having them separate causes issues with Flathub distributions as external release files cause missing release information there. The description field cannot contain list formatting so it was of limited use so the releases are now just links, which also makes this easier to edit.
- Rename the Exult Studio app ID from `info.exult.studio` to `info.exult.exult.studio` for better Flatpak compatibility. In a Flatpak bundle, all desktop files (launcher shortcuts) must start with the main app ID `info.exult.exult`, so for a Flatpak distribution bundling Exult and Studio these files all needed renaming and the icon name changing in the desktop file. This is also the recommended approach in the FreeDesktop spec, though not a requirement.

Other minor changes:

- Fix the missing `type` attribute in the metadata `component` tag which causes a validation error
- Add some more optional data to the metadata file


desktop/info.exult.exult.studio.metainfo.xml
DominusExult, line 1: The Exult Studio documentation ist at https://exult.info/studio.php

fpiesche:
I've also added a [`bootstrap` shell script](https://www.gnu.org/software/automake/faq/autotools-faq.html#What-does-_002e_002fbootstrap-or-_002e_002fautogen_002esh-do_003f) for autotools - this isn't technically necessary but hugely simplifies integration with Flathub's CI.

fpiesche:
bootstrap is just an alternate (and now-recommended) name for autogen.sh. Currently I'm just adding this ad-hoc in the Flatpak build manifest because I had assumed existing builds just involved doing the autoreconf task manually and the bootstrap/autogen script had never existed in the first place. Going to purge that :)

Dragon-Baroque:
> bootstrap is just an alternate (and now-recommended) name for autogen.sh

Yes, I found that in the documentation of the autotools with a discussion about why `bootstrap` was preferred to `autoreconf`. But this document is dated 2015 and reports problems with `autoreconf`.

At this moment, the CI build scripts on GitHub use `autoreconf -v -i` instead of an `autogen.sh` of long ago. I pushed the change. Please see `.github/workflows/ci-linux.yml`. Could the FlatPak builds use `autoreconf -i -v` too ?

fpiesche:
> Could the FlatPak builds use `autoreconf -i -v` too ?

That is in fact what they're doing - Flatpak has support for simplified manifests for a range of build systems such as cmake, ninja, and also autotools, which will handle the actual build commands and common Flatpak-specific build configuration options for these systems transparently. However, `flatpak-builder` expects there to be an `autogen.sh` or `bootstrap` script to be present in the case of autotools; as such what I've done is simply add a `bootstrap` script that runs `autoreconf -i -v` as an ad-hoc patch in the Flatpak manifest.

fpiesche:
Had another look over and tested out a build of the Flatpak manifest with these changes - this should be all that's needed for me to submit Exult to Flathub :)

PR #630, opened 2025-02-17
Permit alternate installation folders for the Shape Thumbnailer configuration files

Dragon-Baroque:
The aim is to remove the need to be root when installing the Shape Thumbnailer as it installs two configuration files into `/usr/share` :
- `configure.ac`
```
      Add an optional target to the --enable-gnome-shp-thumbnailer option into GTHUMB_PREFIX
        system or (default) yes : /usr/share       ( requires root at install )
        local                   : /usr/local/share ( requires root at install )
        user                    : ~/.local/share
        ingame                  : <Exult>/share ( requires adding that folder to XDG_DATA_DIRS )
```
- `tools/x-shapefile.xml`
Increase the priority to 60 to pass over another .SHP file type
- `tools/Makefile.am`
Rework the installation of the configuration files and the `update-mime-database` step using `GTHUMB_PREFIX`

### Note to Reviewers

- The option inventory of `--enable-gnome-shp-thumbnailer` should go into some documentation for Exult, but I have not found a suitable place.
- The increase of the priority over the default ( 50 ) is not nice, I hoped that the Exult thumbnailer would be tested _after_
an installed SHP thumbnailer as this one would fail its magic check on the SHP file, but I could not make that work.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -41,7 +41,8 @@ cxx_compilers="g++ clang++ c++ gpp aCC CC cxx cc++ cl.exe FCC KCC RCC xlC_r xlC"
 
 # determine windowing system from 'host'
 AC_MSG_CHECKING([windowing system])
-HOME_DATADIR="${XDG_CONFIG_HOME:-${HOME}/.config}"
+HOME_CFGDIR="${XDG_CONFIG_HOME:-${HOME}/.config}"
+HOME_SHRDIR="${XDG_DATA_HOME:-${HOME}/.local/share}"
 case "$host_os" in
 	linux-android*)
 		# Android cross compiled from Linux, supported.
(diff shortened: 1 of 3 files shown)


Dragon-Baroque:
@DominusExult, the `ci-macos` workflow builds the Gnome Shp Thumbnailer.

Does it work on MacOS ? Have I disturbed its working with this Pull Request ?

Dragon-Baroque:
Then, if it impacts only the Linux like, do you agree to the Pull Request ?

Dragon-Baroque:
Oh and by the way, the Thumbnailer may be wrongly named as Gnome, for
- It works with the Gnome File Manager `Nautilus` but also with the KDE File Manager `Dolphin`, _could it work with the MacOS Finder_ ?
- It does not depend on any Gnome related include or library,
- It installs its configuration files - that was the point of this Pull Request - into rather generic folders, not precisely Gnome related : `/usr/share/mime` and `/usr/share/thumbnailers`.

I tried to locate which packages own `/usr/share/thumbnailers`, but all the Thumbnailers do, and they do not like Gnome related. A Thumbnailer is actually nothing more than a converter to PNG, with a source file, a target file, and a target size being given in the command line.

marzojr:
When I created it, I only ever tested on Nautilus. I did not know that KDE's file manager used the same architecture. 

And it also used some gnome stuff back then.

Dragon-Baroque:
> And it also used some gnome stuff back then.

It depends only on the `gdk-pixbuf2` which may have split from Gnome proper.

Dragon-Baroque:
> And it also used some gnome stuff back then.

The Gnome dependency ceased by commit 162bfc77.

Not that it is important, should the Thumbnailer be renamed to `exult-shp-thumbnailer` ?

marzojr:
Sure, why not.

PR #633, opened 2025-02-24
Fix double automatic door opening

### LINKED ISSUE #373: Automatic door opening and closing is bugged
In fact, the right mouse button pressed 'movement' behavior of opening doors seems bugged here (test in pc this time).
The avatar opens the door, enters, closes the door, _opens the door again, exits, closes the door the moves inside the room through the closed door_

It's very strange

_Originally posted by @i30817 in https://github.com/exult/exult/issues/356#issuecomment-1593946779_

DominusExult:
Fixes #373

When the Avatar "runs" (via mouse movement) against a closed door, he will open it, go through, turn to the door, close it, turn away. This was bugged as in the Avatar opened and closed it twice, moving through the locked door at times.

### DIFF
--- a/actions.cc
+++ b/actions.cc
@@ -211,6 +211,19 @@ int Path_walking_actor_action::handle_event(Actor* actor) {
 		set_subseq(nullptr);
 		// He was stopped, so restore speed.
 		actor->set_frame_time(speed);
+		if (!door_sequence_complete) {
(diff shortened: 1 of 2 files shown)


DominusExult:
less invasive code as the previous one but I'm not entirely sure it is safe with the use of the door_obj weak pointer. So, please someone review.

PR #637, opened 2025-02-27
Fix compilation when DEBUG_PATHS is defined

vincele:
Error : `src_dir` not defined

Fixing by harmonizing the `src_dir` handling like the other ones.

But I'm not sure this is the right fix, because of the comment on line:
https://github.com/exult/exult/blob/13cb82a544d0b807e9e18f7091479994a0bfee10/gamemgr/modmgr.cc#L184

Please review and advise for the proper fix.

### DIFF
--- a/gamemgr/modmgr.cc
+++ b/gamemgr/modmgr.cc
@@ -512,6 +512,7 @@ ModManager::ModManager(
 	{
 		string patch_dir;
 		string mods_dir;
+		string src_dir;
 		string default_dir;
 		string config_path;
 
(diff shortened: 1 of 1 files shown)


wench (approved):

wench:
This fix looks good to me. I'm not sure the debug output of the source path was ever correct. Prior to the commit that removed the src_did variable, it never even had it's value correctly set. The changes here seem to me to actually be doing what was intended. I'll merge this unless someone else has an objection

PR #648, opened 2025-03-13
Audio changes

### LINKED ISSUE #346: Discover soundfonts in <DATA>/<BUNDLE> and list them in the Audio gump
FluidSynth and CoreAudio need/support soundfonts. Right now you need to either point at them in exult.cfg with the full path or just their filename (if they are in <BUNDLE> or <DATA>).
It would be great if we were to discover all soundfonts in <BUNDLE> or <DATA> and have them as a list you toggle through in an Audio gump setting.
This would mean that you'd need to shorten their name (a lot of soundfonts from the net have ridiculously long names) and recognize if there is a soundfont setup with a full path that is NOT in <DATA> and somehow keep that in the list.

Filetypes useable by FluidSynth are SF2, SF3 and DLS (if FluidSynth was built with SF3 and DLS support); Coreaudio can use SF2 and DLS files.

Space is limited in the Audio gump. 
My proposal would be to make the SFX option into one multiple choice setting: Disabled, SoundBlaster, Roland MT32, MIDI sfx (if enabled).
Probably something that could be done for Speech, too: Disabled, Enabled, With Subtitles.

### LINKED ISSUE #553: U7 Intro - Speech is cut off in new German translation
The new intro speech of the Guardian in my German transslation is longer than the original one. While that is no problem in DOSBOX, Exult cuts it off after the original time slot has ended.

Check here: https://youtu.be/xr2hXNGWZ6Q?t=174
Guardian starts talking right away at 3:07 and will be cut off at 3:57

Any chance to replicate the DOSBOX behaviour and wait for the sound file to have ended?

### LINKED ISSUE #609: Music tracks start abruptly when using non-digital music
When using non-digital music tracks, that is any MIDI driver I've tested including MT32 embedded emulator (MUNT), all tracks start abruptly; more as if they'd start too quickly than missing specific notes. This doesn't happen when using digital OGG tracks.

This issue didn't happen on Exult 1.8.0 but it definitely happens both with 1.10.1 and current 1.11.x snapshots.

### LINKED ISSUE #649: Final Guardian Spech Subtitle line does not show in BG intro
As the title says In the BG intro the subtitle for the final line of Guardian Speechdoes not show. I alredy have this fixed locally. Only really making this issue to document the problem. It will be included in my Audio Changes PR

wench:
A number of Audio related changes 
Primarily to address #553, #346 and more fixes related to #609 

Contains a number of Gump related changes to implement new Advanced Midi settings gump as well as a new drop down widget used by it

Closes #553 
Closes #346 
Closes #609
Closes #649


Dragon-Baroque:
Nice touch to excise the Settings into a separate gump.

This is the first time I can set ( from the Audio gump ) the ALSA server port to my MIDI synth capable Audigy sound card ( in Linux, ALSA is the only way to reach the Audigy hardware MIDI synth ), and this works !

audio/AudioChannel.cc
```diff
@@ -202,6 +206,20 @@ namespace Pentagram {
 		} while (bytes != 0);
 	}
 
+	uint32 AudioChannel::getPlaybackLength() const {
+		if (!sample) {
+			return UINT32_MAX;
+		}
+		return ((sample->getPlaybackLength()) * 1000) / sample->getRate();
```
marzojr, line 213: How likely would this be to overflow?

audio/AudioChannel.cc
```diff
@@ -202,6 +206,20 @@ namespace Pentagram {
...
+	}
+
+	uint32 AudioChannel::getPlaybackPosition() const {
+		if (!sample) {
+			return UINT32_MAX;
+		}
+		return (overall_position * 1000) / sample->getRate();
```
marzojr, line 220: How likely would this be to overflow?

audio/midi_drivers/XMidiEvent.h
```diff
@@ -15,77 +15,173 @@ GNU General Public License for more details.
 
 You should have received a copy of the GNU General Public License
 along with this program; if not, write to the Free Software
-Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.
+Foundation, Inc.,=  59, Temple Place - Suite=  330,, Boston, MA =  02111,-1307,
+USA.
```
marzojr, line 19: There seems to have been some overzealous replacing in this file :)

audio/midi_drivers/XMidiFile.cpp
```diff
@@ -1195,13 +1284,16 @@ int XMidiFile::CreateMT32SystemMessage(
 		const void* data, IDataSource* source) {
 	CreateNewEvent(time);
 	// SysEx status
-	current->status = 0xF0;
+	current->status = int(MidiStatus::Sysex);
```
marzojr, line 1287: Would it make sense to make status be of the enum type instead?

marzojr (approved):

audio/midi_drivers/XMidiEvent.h
marzojr: (this is for the entire license section, but GH was, for some reason, not allowing me to post a comment for the whole range)

Dragon-Baroque:
Although the `DropDown_widget` has the `character_input` handler, no key press event reaches it.
Isn't there some missing `character_input` handler in the `AdvancedOptions_gump` to dispatch the key press events to the children widgets, like there are the `mouse_up`, etc handlers ?

I noticed that while trying to merge this Pull Request with `exult-sdl3`. No problem in the merge, by the way, once I understood that you renamed the `text_input` with three arguments  to `character_input` in the `gumps/` folder.

wench:
> Although the `DropDown_widget` has the `character_input` handler, no key press event reaches it. Isn't there some missing `character_input` handler in the `AdvancedOptions_gump` to dispatch the key press events to the children widgets, like there are the `mouse_up`, etc handlers ?
> 
> I noticed that while trying to merge this Pull Request with `exult-sdl3`. No problem in the merge, by the way, once I understood that you renamed the `text_input` with three arguments to `character_input` in the `gumps/` folder.

Yeah the dispatch should be added.I renamed the function becaise one of the compiler (clang probably) was complaining about overloaded virtuals named text_input
The intent was to allow keyboard input for selecting drop down items but I never really finished what I intended. I should make what's there actually work

audio/midi_drivers/XMidiFile.cpp
wench: Not really, status doesn't just contain valid enum values it is (MidiStatus | channel_number)

audio/midi_drivers/XMidiEvent.h
wench: Yeah I think i did a regex replace and it seems to have mangled the header I'll fix it

audio/AudioChannel.cc
wench: It would probably overflow at the end of the ogg music tracks. More than 2 million samples isn't unexpected for music Will make this 64 bit

wench: Now fixed

audio/midi_drivers/XMidiEvent.h
wench: Fixed the commemts

marzojr (approved):

PR #654, opened 2025-03-16
Exult sdl3

Dragon-Baroque:
With `exult-sdl3` rebased to `master` with the new rework from Wench, try the various builds again.


wench:
Nice to see they are now all passing.

Dragon-Baroque:
> Nice to see they are now all passing.

Yes, thanks to you and to the SDL team, this is one less headache on the path to SDL 3.

Before I claim that all is right, though, I have two worries :
- The OmniOS build of Exult over SDL 3 produces compilation warnings that my own Linux build does not. Quite a number can be removed by bracketing the include of `SDL_keycode.h` in `Gump.h` into the `pragma GCC diagnostic` we use, say, in `exult.cc`.
But that my Linux does not produce them is mysterious.
- I had some failures in making the OmniOS build right which polluted your email boxes.
They boil down to `configure.ac` not recording the CFLAGS of FluidSynth - and probably other packages -, only the LIBS. The OmniOS build installs FluidSynth in `/opt/local`, which is _not_ in the System - permanent - include list.
The past SDL 2 Exult build on OmniOS succeeded because SDL 2 was _also_ installed in `/opt/local` and _its CFLAGS_ is kept in `configure.ac`. Thus FluidSynth was found by a side effect of the SDL 2 CFLAGS.
I finally made Exult SDL 3 build on OmniOS by ensuring that SDL 3 was installed in `/opt/local` too. But this is a hack. I think the `configure.ac` handling of the CFLAGS should be cleaned.

Dragon-Baroque:
@DominusExult I am sorry, the `ci-macos` workflow fails early at the update of Brew with `Failed to download https://formulae.brew.sh/api/cask.jws.json!` which is told to be a transient failure.

I am trying to silence the compilation warnings and I recalled why Linux does not show them.
The reason is that Linux installs SDL include files into `/usr/include` or `/usr/local/include` whereas the other systems install SDL elsewere, into `/opt/local` for example.
`/usr/include` and `/usr/local/include` are considered _System_ include folders and the compilers deactivate some warnings while handling the included files. Not so with `/opt/local`.

I have in this Pull Request that CodeQL has issued some reasonable advice. I shall study this.

Dragon-Baroque:
I would like to merge that PR into master later today, May 14.
Sorry to be rude, but @DominusExult, @marzojr, @wench, do you have any objection ?

wench:
No objections here

marzojr:
No objections.

DominusExult:
Please go ahead!

Issue #571, opened 2024-09-28
List Field tournament issues


mbanck:
I tried to get Dupre's shield back from Luther, but to no avail, I think there are issues with the "fight" mode in the List Field in SI/SS:

1. It looks like the knights never lose, they just flee and then while fleeing, if hit, never get their hit points lowered
2. If other party members fight, they follow the fleeing knights, disregarding other knights that still fight them
3. Eventually they will get hit enough times so that one of the other knights "wins"
4. Non-Avatar party members lose hit points much less easily as the Avatar, so I have the feeling their armor is not being replaced for leather armor like in the case of the Avatar?

This is different in training mode: there, trainers concede being beaten by the trainee.

I did get Luther to say something like "Thou hast beaten me" when he starts fleeing (but just in-game, not as a real conversation overlay); is that the usecode trigger that he will give back Dupre's shield? Because he did not do so next time I met him elsewhere and do seem to recall this would be an immediate conversation on the list field.

mbanck:
Ok, just after I filed the issues, I tried to fight once more, this time with the Avatar. And there, Luther conceded and then subsequently handed me over Dupre's shield on our next conversation.
So it seems only if non-Avatar party members fight the concession check is off?

KnightCaptainU7:
This may match the canon. Luther feels insulted by and angry at the Avatar, not the party, He drops his usual stance when battling the Avatar only as a result.

DominusExult:
I think this can be closed as working as intended.

mbanck-ntap:
While the Luther story may be the same as in the originals, I am pretty sure the way fighting (not training) currently works in exult is different/weird/buggy

DominusExult:
hmm, yes, you are right, I was tunnel visioning the Luther issue. Reopening it.

KnightCaptainU7:
> While the Luther story may be the same as in the originals, I am pretty sure the way fighting (not training) currently works in exult is different/weird/buggy

This is likely due to the fact the the "fight" option doesn't let you pick a particular opponent, just whoever is on the List Field when you go in. Their alignment goes to Chaotic. After the "fight" ends they go back to Neutral.

mbanck:
I think the main problem is that the party member that fights just keeps fighting the fleeing opponent knight instead of picking another one. In some rare cases (probably when the fleeing opponent is far away and some other knights hits them), the switch foes, but usually they just continue to pursue the fleeing knight forever.

[...]

PR #669, opened 2025-04-01
Fix List Field fights

mbanck:
During List Field fighting, the party member fights against all knights. Once the current opponent has been defeated, it does not die but flees as the tournament flag is set. However, the party member has been pursuing this fleeing opponent endlessly, trying to hit it. But even if they hit, they will not make any more damage due to the active tournament flag. Any other knights are ignored and they will eventually win against the party member.

To fix this, clear the target before we call the "died" usecode, so that the party member can select a new foe.

Fixes: #571

### DIFF
--- a/actors.cc
+++ b/actors.cc
@@ -3034,6 +3034,10 @@ int Actor::reduce_health(
 		// No more pushover banes!
 		if (npc) {    // Just to be sure.
 			set_oppressor(npc->get_npc_num());
+			// Unset target, so that we are not endlessly pursuing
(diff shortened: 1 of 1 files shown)


DominusExult:
we really need to run some tests in crucial parts of SI.
Let's collect when tournament flag is important (right now can only think of those):
- Dreamworld fight of Batlin vs the Dream Avatar
- Cantra vs Batlin in the magic globe scene

DominusExult:
I tested some scenes and all seemed to work fine!
Thanks for the fix!

PR #672, opened 2025-04-02
Tools restructure

DominusExult:
The restructuring is looking good now. I've also added mockup and smooth to the autotools.
However for this I had to delete `AC_DISABLE_SHARED([])` from configure.ac and I'm not sure that is not going to pose any problem. However this has been in configure.ac for over 20 years now and probably because it was best practice back then. The plugins of smooth needed this to be gone, though.

I'm also not sure if I need to do something makefile.common (apart from fixing paths as I've done).

Let's let the tests run and I'll go over everything once more to make sure it's ok.

@Dragon-Baroque it's not going to clash with your GTK4 fork, right? And not that much with your SDL3 one?


marzojr (approved):

Dragon-Baroque:
> @Dragon-Baroque it's not going to clash with your GTK4 fork, right? And not that much with your SDL3 one?

I have just made a clone of your fork and I imported my fork into it.

- Rebase of my GTK 4 branch above your `tools_restructure` branch : no problem at all
- Rebase of my SDL 3 branch above your `tools_restructure` branch : some ordinary rework of `configure.ac`, the `Makefile.am` of Smooth and Mockup, and some rework of the `windows-snapshots` workflow, but this last one is not finished, since it depends on our decisions around 32 bits with SDL 3. You may have noticed that I have been able to rework the `exult-sdl3` branch, it now contains six commits instead of twenty plus.

So : no problem from my side.

wench:
Rebase against master for MSVC fix. Turns out it is was just cmake 4.0 compatibility and the latest version of the package fixes it

wench:
Fluidlight and mt32emu also now fail because of Cmake 4 so removing them in MSVC builds for now

PR #674, opened 2025-04-07
Fixes two problems with the two FluidSynth libraries when both are installed

DominusExult:
- in `Makefile.am` both libs were linked
- In `configure.ac` there was no way to actually use libfluidsynth, it defaulted to libfluidlite for `enable-fluidsynth` with both `yes` and `lite`

Now in the case libfluidsynth is not found but libfluidlite is, we will use libfluidlite. Not sure if the same should be done for `enable-fluidsynth=lite`, so if lite is not found we would look for libfluidsynth.


Dragon-Baroque:
I am afraid we need other choices for `--enable-fluidsynth`.
I suggest :
- `lite` : want and use FluidLite, abort if FluidLite is not installed
- `synth` : want and use FluidSynth, abort if FluidSynth is not installed
- `preferlite` : want FluidSynth or FluidLite, use whichever if only one is installed, use FluidLite if both are installed, abort if neither is installed
- 'prefersynth` : want FluidSynth of FluidLite, use whichever if only one is installed, use FluidSynth if both are installed, abort if neither is installed.

Now the tricky one is `--enable-fluidsynth` or its equivalent `--enable-fluidsynth=yes`. Which of the four above does it mean ? One of `preferlite` and `prefersynth`, I guess, but which one ?

Also please restore the two PKG_CHECK_MODULES into the `if test x$enable_fluidsynth = xyes; then`, I moved a few of them in `configure.ac` to check for installed modules only when we enable them.
And please do not place them between the `AC_MSG_CHECKING` and its  `AC_MSG_RESULT`, this produces a confusing log :
```
checking if we want to use fluidsynth midi... checking for fluidsynth... yes
checking for fluidlite... no
yes
```

Dragon-Baroque:
And please not not worry about the FreeBSD build failure. It looks like the `pkg` is missing a few `-y` options.

Dragon-Baroque:
> I'm not sure we really need to handle the edge case of someone having installed both fluidsynth and fluidlite and in the case of configuring one and not finding it, wants to abort?

I am confused whether you mean two cases, or only one. Let me rephrase the matrix.
Assumption : Installed means devel-installed, with includes, pkgconfig, etc. Not runtime only.
Assumption : `lite`, `preferlite`, etc, are the values of the `--enable-fluidsynth` option to `configure.ac`.

1- Only FluidSynth is devel-installed. If `lite` then abort, if `synth` or `yes` or `preferlite` or `prefersynth`, then use FluidSynth and report this in the `can use FluidSynth : yes, FluidSynth variant` ( or simllar ) message.
2- Only FluidLite is devel-installed. If `synth` or `yes` then abort, if `lite` or `preferlite` or `prefersynth`, then use FluidLite and report this in the `can use FluidSynth : yes, FluidLite variant` message.
3- Both are devel-installed. If `synth` or `yes` or `prefersynth`, then use FluidSynth, if `lite` or `preferlite` then use FluidLite and report accordingly.

Which case does look unneeded ?

Please build `FLUID_CFLAGS` like you build `FLUID_LIBS` and use it appropriately in the `Makefile.am`. Please do not assume that the root folder is always part of the standard include set ( `/usr/local/include` and `/usr/include` ). I now force the locally built SDL 3 to `/opt/local` to ensure compliance.

If you want to fix the FreeBSD workflow, the change should be `pkg upgrade` to become `pkg upgrade -y` at line 121 of `.github/workflows/ci-freebsd.yml`.

DominusExult:
> > I'm not sure we really need to handle the edge case of someone having installed both fluidsynth and fluidlite and in the case of configuring one and not finding it, wants to abort?
> 
> I am confused whether you mean two cases, or only one. Let me rephrase the matrix. Assumption : Installed means devel-installed, with includes, pkgconfig, etc. Not runtime only. Assumption : `lite`, `preferlite`, etc, are the values of the `--enable-fluidsynth` option to `configure.ac`.
> 
> 1. Only FluidSynth is devel-installed. If `lite` then abort, if `synth` or `yes` or `preferlite` or `prefersynth`, then use FluidSynth and report this in the `can use FluidSynth : yes, FluidSynth variant` ( or simllar ) message.
> 2. Only FluidLite is devel-installed. If `synth` or `yes` then abort, if `lite` or `preferlite` or `prefersynth`, then use FluidLite and report this in the `can use FluidSynth : yes, FluidLite variant` message.
> 3. Both are devel-installed. If `synth` or `yes` or `prefersynth`, then use FluidSynth, if `lite` or `preferlite` then use FluidLite and report accordingly.
> 
> Which case does look unneeded ?

In my opinion the settings we have now are good enough: default (yes), lite, no

1. Only FluidSynth is devel-installed (the main use case). If `lite` then synth will be used (with a yet, to determined message).
2. Only FluidLite is devel-installed. If enabled without arg or `yes` (so wanting synth explicitly or not knowing that there is another option), then lite will be used (with a yet, to determined message).
3. Both are devel-installed. Either you don't make a decision (synth will be used) or you consciously chose which to use.

IMO, the `prefer...` add a complication that almost no one needs. Most people will just have the lock stock Fluidsynth. If you actually know about Fluidlite and installed the dev package (is it even in any package manager?) you are likely to explicitly want to use it.
Aborting configure if we have either (but configure was set to the missing), is a hurdle that we can just avoid. OR we just don't offer the fallback option if we rather want to abort.

> Please build `FLUID_CFLAGS` like you build `FLUID_LIBS` and use it appropriately in the `Makefile.am`. Please do not assume that the root folder is always part of the standard include set ( `/usr/local/include` and `/usr/include` ). I now force the locally built SDL 3 to `/opt/local` to ensure compliance.

Good point about the `FLUID_CFLAGS` - I only need to do the same as for `FLUID_LIBS`, just with CFLAGS in audio/midi_drivers/Makefile.am. Will do tomorrow.
On macOS only the Apple provided developer stuff is in /usr/include and you are not supposed to pollute /usr/local/include, so I'm always using a PREFIX in a custom location.
 
> If you want to fix the FreeBSD workflow, the change should be `pkg upgrade` to become `pkg upgrade -y` at line 121 of `.github/workflows/ci-freebsd.yml`.

thank you!

Dragon-Baroque:
Looks all good to me, passes all builds. Merge ?

PR #680, opened 2025-04-11
Fixes #679

### LINKED ISSUE #679: SI: Polymorphed NPCs's dead bodies should revert to their real bodies
Before the wall of light you are encountering Selena and Batlin's other henchmen. Selena polymorphs them into Cyclops.
In the original, when you kill them, they revert to their real bodies. In Exult they stay cyclops'.

Savegame in Exult just before the encounter (go north):

[exult237si.sav.txt](https://github.com/user-attachments/files/19714611/exult237si.sav.txt)

Original savegame:

[GAME09.U7.txt](https://github.com/user-attachments/files/19714613/GAME09.U7.txt)

DominusExult:
Dead polymorphed NPCs should get their real dead body

### DIFF
--- a/actors.cc
+++ b/actors.cc
@@ -4415,6 +4415,14 @@ void Actor::die(Game_object* attacker) {
 	// Get location.
 	const Tile_coord pos = get_tile();
 	// properties[static_cast<int>(health)] = -50;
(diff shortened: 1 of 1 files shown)


DominusExult:
Can anyone think of a situation where this could be problematic?

PR #688, opened 2025-04-29
Optimise svg icons and pngcrush Exult Studio png

### LINKED ISSUE #671: The files of the icons info.exult.exult(.studio).svg are big
- `info.exult.exult.svg` is 270 K, `info.exult.exult.studio.svg` is 210 K, mostly because they contain a 16 K javascipt script, replicated 8 times in `info.exult.exult.studio.svg`, replicated 15 times in `info.exult.exult.svg`.

If I remove the scripts, I get a seemingly identical image, but 14 K for `info.exult.exult.svg` :

Image

and 92K for `info.exult.exult.studio.svg` :

Image

@fpiesche, would you like to validate the reduced size icons above ? 
Just right click on them and Save as, as they are the SVG I obtained after slicing off the scripts, just renamed.

fpiesche:
Fixes #671. Removed the unnecessary Javascript from the svgs and applied `pngcrush` to all the png exports while I was in there. The SVGs are now 48kB (Studio) and 14kB (Exult) respectively, making them smaller than the 512x png exports. As it should be :)

From the look of the Javascript in there, it was probably left over in the files from my initial attempts at creating the metallic shading on the ankh with mesh gradients - however these aren't a standard SVG feature yet so wouldn't render at all in web views (ie on Github) or when used as icons, so I redid them with multiple separate shapes with linear or radial gradients instead. Apparently Inkscape wasn't smart enough to also remove the Javascript defining the mesh gradients when I switched the fill mode, even when saving as "Optimised" SVG.

I've also fixed some minor precision issues on the ankh in both versions that were visible at high zoom levels - some vertices that should have been in the same place weren't, so there were some tiny gaps along some of the edges and some of the corners hadn't had their edges converging on the same point properly.


DominusExult:
thanks. Any idea why it says there are conflicts with the png files?

wench:
I'd guess the conflicts are because the branch hasn't been rebased to master and is from before the latest run of the PNG optimise bot

fpiesche:
oop, my bad, will sort that out after dinner!

PR #695, opened 2025-05-07
Fix Studio icon

fpiesche:
Looks like this was a copy&paste error - I'd just copy&pasted the ankh on the previous PR but must have missed out some component.

I've now used Inkscape's import functionality to import the fixed base icon into the Studio svg, repositioned it where it belongs and deleted the old version. Checked against a magic pink background and looking good.

image


fpiesche:
... what in the WORLD

DominusExult:
I've been working with svg files for the last two days and I've been mixing them up left and right :)
Tried to do a nice logo for our website and ran against stupid problems caused by Safari:
imbedding an svg like an image is no problem but if you use
```
<svg>
<use href="path_to_file"></use>
</svg>
```
Safari suddenly can't handle `filter:"url(#xyz)"` anymore 🤦‍♂️
Took me a while to accept this and now we just embedd the svg 🤷‍♂️

fpiesche:
... for some reason the regular ankh icon on `master` also seems to be lacking the background ankh. Hnnnng

DominusExult:
huh? that's odd, I can clearly see the background ankh in our master, it's `path66` in group `g69`

fpiesche:
Maybe I'd branched off too early. Now double checked that they both have a background ankh, and also the gaps fixed otherwise so the background ankh *shouldn't* be necessary but hey ho, it'll cover up any other precision problems we might run into.

DominusExult:
I'm really sorry for bearing bad news, especially after all you've done already. 
I noticed something was off in the exported pngs: your current ankh backgrounds are just barely overlapping the foreground everywhere. it's also present in the ankh svg from the accepted PR from a few hours ago but since it's now two layers of this, it's more apparent.
I painted all the backgrounds in group 78 red and got this:
image

Of course the red makes it stand out more than it actually looks like but it shows the problem.
I also can't find any fault anymore without the background, so maybe you can really abandon it.

If you are fed up with this now, I can do that. I would understand this completely. Doing one thing just to get Exult on flatpack and it haunts you for months :)

DominusExult:
@fpiesche I've now gone ahead with just removing the ankh background, same for the ankh in the Studio icon, updated all files that use these. Looks good now. So I'm closing this PR.
Thanks a lot for doing this. I really liked the icons you did before but working with these for the last few days, I appreciate them even more!

PR #697, opened 2025-05-10
Windows crash handler

wench:
Here is my Windows Crash Handler changes. This PR is here to document the changes. I;ve split this up into different sections

#### Windows Crash Handler 
Primarily it is intended to make it easier to tell what happened when a use experiences a crash. THe crash handler does the following

1. Output some basic details about the crash. Previously we couldn't tell If Exult even crashed as nothing would be written to stdout or stderr
2. Generate a mini-dump file that can be loaded into a debugger for postmortem debugging (more about this below)
3. Call verify_files  to verify the game files. Just in case that is the problem
4. Try to generate an "Emergency save" to save the game state when the crash occurred 

THe crash handler will be called for the usual hardware exceptions like Access Violations and Integer divide by zero.
I also set things up that it will be called after the Runtime calls terminate or abort. 

#### exult_exception changes
It took the opportunity to update the exult_exception class and how the top level catch block for it behaves
The class now has fields for source file and line That will be set when Constructing sub classes using macros I added
The top level catch block will now output the source file and line for any exceptions it catches. It will also call verify_files and produce an emergency save

#### Verify files changes
Previously verify_files required being passed a BaseGameInfo pointerto the game to checked. I updated it so if passed nullptr it will verify the current game instead. I also updated it so files canhave an Alternative hash if there are different versions ofthe File I the wild

I noticed that my BG USECODE file would verify bad. I realized that is because there are 2 different version of English BG and the Only Datafile difference is in The Usecode. Exult's hash was for the V 3.4 USECODE file while I have a V 3.0 USECODE file so I added the hash for 3.0 as well. I am curious what the difference is but that is something to investigate at another time

#### Emergency Saves
Emergency saves are a last chance attempt to create a save of the game state Exult was in when it crashed The code is setup to use a temporary GAMEDAT folder so it doesn't overwrite the existing GAMEDAT just incase the Emergency save ends up unusable or Exult crashes while making it. After saving to the temprary gamedat it write a normal savegame file with the name "Crash Save" to the next free save game slot. The code imposes a limit of 1000 save games, but I think this is safe. The Emergebcy save code disables the parts of the Normal save code that updates the screen to print the "Saving Game"message as well as suppresses creating the mini screenshot to reduce what parts of code need to be working properly in case the entire Exult process has become unstable.

If Exult crashes while making an Emergency save it should not cause harm beyond creating a bad save file

#### Mini-dumps
stderr will indicate it created a mini-dump and where it was saved (in the exult home directory)
By default Exult creates a "normal" mini-dump suitable for obtaining a stack back trace however I added a config setting to allow creating a full mini-dump that contains the entire process state, Normal mini-dumps should be about 100 KB while Full Minidumps should be around 200 MB. To get full minidumps 
set config setting "config/debugging/windows/minidumptype" to full

With a suitable version of gdb, the minidumps can be loaded and a stack trace can be obtained. The normal versions of gdb do not support loading windows minidump but the versions of gdb found here: https://github.com/ssbssa/gdb/releases do support loading them. I've been using the latest version [16.3.90.20250420](https://github.com/ssbssa/gdb/releases/tag/gdb-16.3.90.20250420)

To load the minidump. Load gdb with a command like the following
`./gdb64.exe  ./Exult.exe  C:/Users/Colourless/AppData/Local/Exult/minidump.dmp`
You should use the same arch of gdb as the version of exult being debugged. The exult.exe should be one with at least -g1 debugging symbols but it doesn't need to be the exact exe that crashed. If the exe that crashed was the stripped version of the exe being debugged then all is fine.

#### Windows Snapshot workflow Changes
In order to make it easier to debug a minidump from users of the snapshots I made it so the snapshot exe files are built with DEBUG_LEVEL=-g1 The dist makefile targets strip the exes of Debugging symbol when copying them to the installer directory so there should be no size difference for the executables included in the installer.

I added a step to the workflow so the unstripped executables suitable for use with GDB are uploaded as an artifact. The github default keeps artifacts for 90 days.

#### final remarks
Any informat ion that gets output I am writing to stderr

In order to debug a minidump from a user you'll want the unstripped executable from the action artifacts of the run that produced the snapshot and use that as the exult.exe you pass to gdb

The minidumps created after a SIGABORT in Mingw32 builds do not create good backtraces but everything is fine in Ming64 and Mingw Clang64 builds. So I left thiscode in there as we are planing on ending mingw32 builds soon anyway.

Unhandled c++ exceptions thrown by the runtime will get caught by my crash handler and generate mini-dumps

The code I've written here mostly assumes that Exult will still be in a somewhat stable state after crashing. This may not be a reasonable assumption to make so A crush during the crash handler will lead to immediate process Termination


marzojr (approved):
Just needs some clang-format, but looks otherwise.

PR #702, opened 2025-05-14
Better error handling in parsing keybindings

DominusExult:
The keybinding parser should have printed out line numbers on errors but didn't. Now it prints the line number and the offending line itself.
Also fixes a problem when parsing pathkeys: empty lines were no longer skipped.


Dragon-Baroque:
Tested it applied to exult-sdl3. Some easy fixes, please do not worry.
Used a damaged `patchkeys.txt` for the test, the warnings are ok with the offending line number.
Please merge.

PR #716, opened 2025-05-30
Fix SVG attributes for raw display

fpiesche:
For some reason, the presence of the `xml:space="preserve"` attribute on the main `svg` tag in these two images broke rendering of the files in a browser: https://raw.githubusercontent.com/exult/exult/3ecbcdcc181a9edca13d3201875c04dfc7c9163f/desktop/info.exult.exult.studio.svg

This fixes it: https://raw.githubusercontent.com/fpiesche/exult/622d4d30f7eb853f83cee03dc45943dfe0c5a2d1/desktop/info.exult.exult.studio.svg

I've also added the xmlns attributes for good measure. I don't think they're malformed without them but the consistency is nice to have.


DominusExult:
thank you!

PR #749, opened 2025-07-11
Mods override skip_splash and force_digital_music

DominusExult:
mods can override the skip_splash setting (mod_info/skip_splash) to disable the intro in case the mod is a total conversion that has nothing to do with the original game anymore.

There is also the option to force digital music (mod_info/force_digital_music) which is not doing anything yet.

### DIFF
--- a/exult.cc
+++ b/exult.cc
@@ -1051,12 +1051,27 @@ static void Init() {
 		bool skip_splash;
 		config->value("config/gameplay/skip_splash", skip_splash);
 
+		// Check for mod overriding skip_splash
+		bool force_skip_splash = false;
+		if (gamemanager) {
+			ModManager* current_game_mgr
+					= gamemanager->find_game(Game::get_gametitle());
+			if (current_game_mgr && !Game::get_modtitle().empty()) {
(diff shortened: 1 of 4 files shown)


wench:
You don't need to change the driver. You'd want to put you override in the MyMidiPlayer in Midi.h/cc

DominusExult:
What I mean is that Exult could be configured to use FMOpl, so it starts up with that. But when you select a mod that forces digital music, it needs to switch.

DominusExult:
Neat, thanks. I'll get to it tomorrow.

wench (approved):
One minor problem with this that probably doesn't matter, the player could go into Audio settings and Turn off digitalmusic

DominusExult:
That's true, I've also thought of that. Maybe I need to add a check to the Audio gump that warns that the mod is forcing this (or doesn't allow disabling in this case)

wench:
Yeah probably a good idea but if you can wait a few days it would be nice as my mod installation changes ideally would accompany a version bump too so I can detect if exult.exe actually has my changes. I plan to make a PR for it in a few days. so upping the version after then would be best

DominusExult:
makes sense :)

PR #751, opened 2025-07-17
Integrated installers

### LINKED ISSUE #746: Update Windows installer so unzipping files can be done in Wine
Windows installer uses the Windows shell to unzip downloads and mods. This doesn't work in Wine and I should implement a different solution to unzip the files for the small number of users who may want to use Exult in Wine.  mentioned in #745

wench:
This PR is to add integrated mod and data installers into Exult itself.

This primary drive was to remove the need to use Windows shell code in our Windows Installer to unzip the mods and data that would not work in wine.
closes #746 

2 new command line options are added `--installdata <zipfile>` and `--installmod <zipfile>`

`--insatalldata` is quite simple, all it does is extract the zip file argument into `<data>`

`--installmod` does quit a bit more, it looks in the root of the zip for modcfg files.  For each modcfg found, it checks Required_version and prints a warning if the mod is incompatible, but it will still install the mod as Exult itself Shouldn't show incompatible mods in the menu. It will then attempt to figure out what game to install the mod to by getting the identity from the mods initgame file if it has one. For mods that do not change the map and so do not have an initgame, it will attempt to get the identity from a new modcfg field game_identity.  After identifying the game, installmod will delete any existing mod patch files if there were any and then extract the mods data from this zip. When extracting the files, the process will rename the mod patch files so they match the case of the files in `<static>`. --installmod can be combined with `--bg` `--fov` `--si` `--ss` or `--game` to force choose what game to install a mod to overriding auto detection using identity. It will print a warning if the user tries to force installation of a mod into an incompatible game but will allow it. Force choosing the game is required if the mod does not have an initgame and does not have game_identity set (this would be needed for graphics only mods that can be used with any game). The install mod code can install multiple mods from a single zip file provided they are all in the root of the zip with a unique name and are well formed. If there are multiple mods they can all be for different games, so a single zip could contain multiple different versions of the same mod for different games, or be a mod pack of completely unrelated mods

These changes bumps the version to 1.31.1 Partly because of the recent changes by @DominusExult  to modcfg as well as to give me the ability to use the exult version number in the windows installer to detect if installmod and installdata are support

In order to make game_identity  in modcfg somewhat useful, I altered Exult studio so when creating  a new mod it appropriately sets game_identity. I also took the opportunity to have Exult_studio save out default values for some of the other modcfg fields that modders might find useful. It now also saves defaults for `patch`, `source` and the recently added fields `skip_splash`, `clean_menu` and `force_digital_music` as we don't really document any of this stuff anywhere

I lastly updated the Windows installer to use --installdata and --installmod.  In the event that the user has selected to install mods or data but has deselected to install exult, the installer will check the version of the already installed version of Exult. If it is older than 1.13.1 or wasn't installed, the installer will fallback to the old Windows shell method to extract the zips. InnoSetup 6.4 is now required as it seems I am using a scripting function that was added in 6.4

The windows installer now has a few important behavioral changes. Firstly it will now report errors if the user tries to install the mods and they do not have the expansions. All our mods require the expansions, previously the installer would install the mods and Exult would fail when trying to start a new game in a mod because the identities didn't match. This also makes sure the mods get installed to the correct games if you have all 4 games version installed. Previously if you had both the  expanded and unexpanded games, the old Windows installer code would install the FoV mods to the `<blackgate>` and SS mods to `<serpentisle>` regardless of what games they actually are. Now Exult installs mods to the correct game. Secondly if the user has made a mod file read only the new installation procedure will get an error while the old shell method would silently  overwrite read only files. Mods and data can now be installed in wine provided Exult is also being installed or the existing exult install is 1.13.1 or newer.

This also makes makes quite a few changes to the unzip code. Instead of opening a zip file from afilename and using a FILE* the unzip code now uses IDataSource* also to read zip files store in a memory buffer. This change was done so I could load a mods initgame into memory and read identity from it and initgame will most likely be a zip file and I didn't want to write out initgame to a temporary file just so the unzip code could read it to get the identity.

Changed the `get_game_identity` function to take an IDataSource* and added an overload with the old filename argument to maintain compatibility with existing code

I split off the code that checks modcfg required_version compatibility out of the `ModInfo` constructor into it's own static method so I could use it in the mod Installer without needing to duplicate code

`ModManager` now stores `static_identity` in a member variable 

`--installmod` and `--installdata` should be usable on any operating system though for most users they would need to be run as root or Administrator. No idea how it would interact with a containerized version of Exult like a flatpak

Made a few changes to clean up `stderr` output so when using `--installmod` and `--installdata` `stderr` will only contain fatal errors not benign errors or other informational messages

### DIFF
--- a/.github/workflows/snapshots.yml
+++ b/.github/workflows/snapshots.yml
@@ -43,7 +43,7 @@ jobs:
             echo HAVE_ANDROID_SNAPSHOT=false >> $GITHUB_ENV
             echo Couldn\'t find exult-snapshot-signed.apk for Android snapshot. No release will be made
           fi
-          echo "GITHUB_REF=v1.13.0.$(date --utc +'%Y%m%d')" >> $GITHUB_ENV
+          echo "GITHUB_REF=v1.13.1.$(date --utc +'%Y%m%d')" >> $GITHUB_ENV
       - name: Install SSH key
         if: ${{ (env.HAVE_WINDOWS_SNAPSHOT == 'true' && env.HAVE_ANDROID_SNAPSHOT == 'true')}}
         uses: shimataro/ssh-key-action@v2
--- a/Makefile.common
+++ b/Makefile.common
@@ -1,4 +1,4 @@
-VERSION:=1.13.0git
+VERSION:=1.13.1git
 
 MAIN_OBJS:=actions.o \
 	actorio.o \
--- a/configure.ac
+++ b/configure.ac
@@ -1,6 +1,6 @@
 dnl Process this file with autoconf to produce a configure script.
 AC_PREREQ([2.69])
-AC_INIT([Exult],[1.13.0git],[],[exult],[https://exult.info/])
+AC_INIT([Exult],[1.13.1git],[],[exult],[https://exult.info/])
 AC_CONFIG_SRCDIR([exult.cc])
 
 # ---------------------------------------------------------------------
--- a/exult.cc
+++ b/exult.cc
@@ -71,6 +71,9 @@
 #include "utils.h"
 #include "verify.h"
 #include "version.h"
+#ifdef HAVE_ZIP_SUPPORT
+#	include "files/zip/unzip.h"
+#endif
 
 #include <cctype>
 #include <cmath>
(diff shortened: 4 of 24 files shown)


wench:
> The "©" in "Copyright © 2024" is damaged in `win32/exconfig.rc`. It was damaged before this PR, mind, but the PR altered it.
> 
> You could fix it on the fly by retrieving the whole copyright statement of either `win32/exultico.rc` or `win32/exultstudio.ico` : `VALUE "LegalCopyright", "Copyright © 1998-2025\0"`.

Fixed

PR #754, opened 2025-07-18
Create scripted cutscenes

DominusExult:
First of full disclosure, this PR was done with lots of help by Github Copilot and this shows in the way the code is constructed.

Playscene allows creating scripted cutscenes using the media Exult knows how to handle.
Namely FLIC animations, audio files OGG, WAV, VOC and also playback of a games'
music and sfx tracks.

To start a scripted scene use
```
if (scene_available("name")) {
			play_scene("name");
		}
```

`scene_available("name")` looks for the files name_info.txt and name.flx in
a game's or a mod's patch folder, which play_scene("name") will play back.

Exult will automatically look for intro, endgame, quotes and credits
cutscenes in the above mentioned folders (files: intro_info.txt,
intro.flx, endgame_info.txt, ...) and play these instead of the orignal.

The script is loosely based on the format we use for de-hardcoded shape
informations.
You can have several scene and text sections which will be played in the 
written order.

docs/playscene.txt has lots more details on how a cutscene is made and at https://youtu.be/JSmtgFyt71c you can see a video of how it plays (that was a bit earlier in the stages of development).

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -99,6 +99,8 @@ EXULTSOURCES =	\
 	paths.cc	\
 	paths.h		\
 	headers/pent_include.h	\
+	playscene.cc	\
+	playscene.h	\
 	readnpcs.cc	\
 	rect.h		\
 	schedule.cc	\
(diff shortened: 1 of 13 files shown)


DominusExult:
> use of designated initializers requires at least '/std:c++20' 

ok, that is something to fix, tomorrow.

PR #756, opened 2025-07-20
Adding play_scene intrinsic.

DominusExult:
With this you can start a scripted scene during the game. It will save the palette and restore it afterwards as fade outs in scenes really messes with the current palette. Additionally the queue is suspended during playback.

### DIFF
--- a/docs/exult_intrinsics.txt
+++ b/docs/exult_intrinsics.txt
@@ -3237,6 +3237,12 @@ obj->obj_sprite_effect(int sprite, int rel_x, int rel_y, int vel_x, int vel_y,
               the sprite frame will go through all of its frames. If negative,
               the sprite will go through all of its frames *once*.
 
+string UI_play_scene(string text)
+    Plays scripted cutscene after checking whether the files text_info.txt and 
+    text.flx exist in the patch folder.
+  Parameters:
+    text    The base name of the cutscene files.
+
 ================================================================================
 OTHER INTRINSICS
 ================================================================================
--- a/usecode/intrinsics.cc
+++ b/usecode/intrinsics.cc
@@ -1,5 +1,5 @@
 /*
- *  Copyright (C) 2000-2022  The Exult Team
+ *  Copyright (C) 2000-2025  The Exult Team
  *
  *  This program is free software; you can redistribute it and/or modify
  *  it under the terms of the GNU General Public License as published by
@@ -50,6 +50,7 @@
 #include "mouse.h"
 #include "palette.h"
 #include "party.h"
+#include "playscene.h"
 #include "ready.h"
 #include "rect.h"
 #include "schedule.h"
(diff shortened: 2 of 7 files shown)


DominusExult:
@marzojr could you review this please? I think I have done everything correct concerning adding the intrinsic but I really don't want to mess it up :)

docs/exult_intrinsics.txt
```diff
@@ -3237,6 +3237,12 @@ obj->obj_sprite_effect(int sprite, int rel_x, int rel_y, int vel_x, int vel_y,
               the sprite frame will go through all of its frames. If negative,
               the sprite will go through all of its frames *once*.
 
+string UI_play_scene(string text)
```
marzojr, line 3240: Please note that this file is generated. In the exult-web repository, I add a "header" for the intrinsic on `seventowers/usecode/intrinsics/defs`, add the documentation body for the intrinsic on `seventowers/usecode/intrinsics`, add the intrinsic under an appropriate group on `seventowers/usecode/reference/intrinsic_relations.dat` (including adding links to related intrinsics), and add it to `seventowers/geshi/ucc.php` for syntax highlighting.

I then generate the doc for Exult through

```bash
$ cd /path/to/exult-web/seventowers
$ php code.php TITLE_IMAGE=usecodetitle.png DATAFILE=exult_intrinsics.dat TYPE=1 OUTPUT=text > /path/to/exult/docs/exult_intrinsics.txt
```

marzojr, line 3240: Also, the intrinsic as defined in [usecode/intrinsics.cc](https://github.com/exult/exult/pull/756/files#diff-36c9a429bcde9092290a67190351652dcb6ba679aac1ab53918bbd969411795a) does not return anything, and so should be `void` return type.

usecode/intrinsics.cc
```diff
@@ -3058,6 +3059,35 @@ USECODE_INTRINSIC(play_sound_effect2) {
...
+		// fade outs in scenes will screw up the palette, so we save current
+		// palette index and restore it after the scene is played.
+		int prev_palette = gwin->get_pal()->get_palette_index();
+		// Pause the queue.
+		gwin->get_tqueue()->pause(Game::get_ticks());
+		play_scene(scene_name);
+
+		// Restore the palette index.
+		gwin->get_pal()->load(PALETTES_FLX, PATCH_PALETTES, prev_palette);
+		gwin->get_pal()->apply(true);
```
marzojr, line 3083: Maybe fading to black before the scene, following by fading the cut-scene to black, then fading from black to old palette? This could even be another parameter to opt in/out of this fading.

marzojr (changes requested):

usecode/intrinsics.cc
DominusExult: Thanks, I'll do the txt file as you wrote and will try with the palette as you suggested

DominusExult: As the text scenes change the palette, I need to keep track of it, but I moved this part off to playscene.cc. Fade_out/fade_in added anyway as optional bools and autogenerated exult_intrinsics.txt. By chance I stumbled over the weird text in teleport_to_saved_pos and corrected that, too. Saves me another commit later on, but I can keep it clean. The seventowers web file changes are lined up as well.

marzojr (approved):

PR #768, opened 2025-08-12
Hide UI elements (face stats, shortcuts bar, virtual joystick) during the first scenes in BG and SI

DominusExult:
Hide UI elements (face stats, shortcuts bar, virtual joystick) during the first scenes in BG and SI.
When the respective global flags are set (0x3b in BG and 0x03 in SI) these are shown if configured to.

### DIFF
--- a/exult.cc
+++ b/exult.cc
@@ -687,8 +687,9 @@ int exult_main(const char* runpath) {
 		touchui->showButtonControls();
 		// TODO(Marzo): This is probably the wrong place for this
 		Usecode_machine* usecode = Game_window::get_instance()->get_usecode();
-		if (!usecode->get_global_flag(Usecode_machine::did_first_scene)
-			&& GAME_BG) {
+		if ((GAME_BG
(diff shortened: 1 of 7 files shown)


wench:
Seems reasonable but one minor concerning modded games changeinhthe first scene and do not use the same flags as the base game the UI elements will never be shown. This change needs to be documented so anyone making a TC mod knows they need to set the correct flags in their first scene usecode

DominusExult:
I've added the suggestions except for 

> I would prefer it if UI elements were hidden based on some non game specific state like avatar don't move being set or something else.Advantage of using something like the don't move flag is that modders can use it to hide the ui during scripted scenes

that's a completely different approach and needs code to hide the Gumps.

But I'll experiment further if this suggestion works and most imprtantly if I can make it work. Something like gwin->main_actor_dont_move() && !gwin->main_actor_can_act() but I need to make sure the UI stays on when the Avatar is charmed (even when !gwin->main_actor_can_act_charmed() as this is very likely not a scripted scene then).

PR #777, opened 2025-09-08
Hide UI during SI's scripted crystal ball scenes.

DominusExult:
Also close all gumps as these were shown during the scenes if opened before.

This works for the Batlin/Cantra and the planetary (cat) scene. No adverse effects on other scripted SI scenes (tested wall of lights, Petra body switch), though these still show the UI.
If you have any idea what to test, additionally, please let me know.


DominusExult:
This works, but I found that if I use `set_item_flag` `dont_move` and `clear_item_flag` to toggle this, I get the same result in crystal ball but all other scripted cutscenes work as well.
I'll do another pull request.

Issue #781, opened 2025-09-11
Serpent Isle intro crashes when attempting to load speech (Spanish version)


AlWesker:
I'm getting this error with the Serpent Isle intro (Spanish version) in the last v1.13.x Snapshot (because with Exult v1.12.0 I can't get to boot neither Black Gate nor Serpent Isle). It happens just before it's going to start the speech.

<img width="640" height="511" alt="Image" src="https://github.com/user-attachments/assets/864da96c-c738-4cf1-9b08-36c38a0fd888" />

If I skip the intro, the game behaves normally, but would like to know how can I solve this. Black Gate (also Spanish version) has no issues and plays both the intro and the game normally.

AlWesker:
It's true. I just tested the last v1.13.x Snapshot and the intro doesn't crash any longer but it doesn't display text. Will that be possible to be solved too?

wench:
Couldn't let this go and had to fix it before bed tonight.

AlWesker:
Thanks for fixing the text issue but there's still a little problem. It's true that the text is now showing in the intro cutscene, but it's in English, when it should be in Spanish as it shows using the original .exe.

<img width="640" height="425" alt="Image" src="https://github.com/user-attachments/assets/65985214-e4dd-489d-8c8b-838d944cf02c" />

<img width="640" height="426" alt="Image" src="https://github.com/user-attachments/assets/108b0550-79c1-4685-97f4-f03b4fa74f69" />

Could this be looked into?

PR #795, opened 2025-09-26
Load translated exultmsg and update gumps and exultmenu so they can be translated

wench:
After the comments in #781 I decided to do something about exultmsg translations not being loaded
I only intended to add support for loading translated exultmsg but got a bit carried away with all this

Firstly Translated exultmsg will be now loaded. By default it will load the exultmsg that matches the language of the game. The language can be overridden globally with <config/gameplay/language> setting it to de/en/es/fr. Additionally language can be set on a per game basis using <config/disk/game/<GAME>/language> The per game setting will override the global setting.  Exult Studio will load the Exultmsg in the language of the game but this shouldn't actually have any effect as Exult Studio doesn't use any of the strings in Exultmsg as best as I can tell

I added selecting the Global setting to GameDisplayOptions_gump. 

I made some important changes to how erxultmsg is loaded. Translation and Patches are loaded ontop of the default english exultmsg, they do not replace it. Missing strings in a patch or translation will fallback to the default english string instead of just resulting in empty strings. This specifically allows partial translations to still work. The French and German exultmsg are missing all the SI specific strings so this was essential now that the translated files will be loaded if the users chooses french or german and plays SI when the strings are missing.

While adding the config option to GameDisplayOptions_gump I decided I should update the gumps so they are all Translatable
and then if I update the Gumps I should also update the small number of string in the exultmenu so they are also translatable.
All the gump string have been moved to exultmsg. The string in exultmsg are separated into different groups for each gumps with a section for shared strings at the top. There is one gump that does not really support Translations and that is The Notebook Gump, in particular we don't current have a way to translate autonotes. The few none autonote strings in that gump (DAY, am, pm) however are translatable. Some scaler names are now translatable for display in VideoOptions Gump but most aren't as they don't have proper word names. Midi driver names used by Audio Options are not translatable as these don't seem to me like things that should be translated from technical POV and I didn't really want to be changing the midi driver interface to add support for translatable display names. The Display Strings used for the Advanced Midi Settings can be translated as many of these likely would have reasonable translations if someone were to translate them and very little needed to be changed to support translations .

While making the gump tanslatable I realized that the fixed horizontal layouts with fixed columns in the colx arrays  wont work if the translated strings are longer than the English ones. So I went through and updated all the Gumps so they have a dynamic horizontal layout and will automatically resize the gumps as needed.

The way the gump layout works is as follows:
1. The Setting Buttons are placed to the right of the label string for that setting.
2.  The gump is resized horizontally to fit the buttons in case any of them are over the edge.
3.  The bottom row of buttons OK/APPLY,HELP,CANCEL, are automatically arranged by evenly spacing them across the gumps, resizing the as needed if the total width of the 3 buttons is wider than the gump. 
4. After resizing the gump to fit, the settings buttons are then all right aligned 4 pixels from the edge of the gump.

With the Horizontal layouts now done automatically I got rid of the colx arrays, With colx gone I also decided to see if I could get rid of rowy too. I replaced the rowy arrays with a conexpr method in Modal_gump that algorithmically calculate the rows. Staring from an initial offset of 5 pixels, Rows 0 to 12 are 12 pixels high and rows above 12 are 14 pixels high. This seems to match what was in the rowy arrays.

Visually the changes to the GUMP layouts are fairly indistinguishable to how things were before these changes. Might not be pixel identical but are close enough IMO

I did not add any translated strings in any of the exultmsg_XX files. I do not know these languages in order to make translations myself and Doing an Automatic translation online without me being able to QC to the result seemed like a bad idea. Using Google translate to translate lines from exultmsg generally seems to work ok so it would be fairly easy to add automatically translated strings in the future if so desired


DominusExult:
wow, quite a rabbit hole you fell into. This looks beyond useful and being able to translate Exult itself is awesome. Thank you

wench:
Will fix ingame menu alignment problems,

German and French fonts might be taller to account for diacritical marks

wench:
Fixed Gamemenu_gump and changed it to autosize its height to match the buttons shown so it no longer has the gap at the bottom when going into settings from Exult menu or in ios

DominusExult:
looking good now.

DominusExult:
Coverity complains about some of your code in this commit and I noticed that you are not signed up for it. Should I send you an invite to the address I've sent you the Spanish files?

wench:
> Coverity complains about some of your code in this commit and I noticed that you are not signed up for it. Should I send you an invite to the address I've sent you the Spanish files?

Ok sure

DominusExult:
Thanks, it's quite a bit, though some are complains about security of rand or getting thrown off because we use framenum 0.

DominusExult:
I only noticed because AgentOrangeguy asked me the exact mod settings to disable menu items and I started his Ultima VI mod game.
I was expecting it to eventually crash as the strings shown weren’t always the same.
Thanks!

DominusExult:
I found one more problem gump:
When audio is disabled the audio options gump is too small for the buttons. To see it, audio has to be disabled before you open the gump. So open Audio Options, disable audio, close the gump and reopen Audio Options.
image

DominusExult:
And toggling through the scalers all scalers after bilinear plus are labeled `"Step aside!"`.
image

DominusExult:
> And toggling through the scalers all scalers after bilinear plus are labeled `"Step aside!"`. 

Ah, I see it's because the other scalers aren't named in exultmsg.txt 🤷

wench:
uh oops scaler names now fixed

DominusExult:
Sorry, playing around with FluidSynth on iOS I found another fault (general, not only on iOS):
image

Enable FMOpl or Fluidsynth and then enable digital music -> 4 rows in the Volume Mixer and the right side goes out of the gump.

DominusExult:
yes, you are right, the culprit is 585211736 which fixed the Gamemenu_gump always resizing bigger after leaving GameDisplayOptions_gump.

PR #828, opened 2025-12-14
Show 3d BBox in ES ShapeInfo editor and Exult

### LINKED ISSUE #830: [Feature Request] Map mode to show 3d boundary boxesd on all shapes at once
I was wondering if something like a feature in Pentagram could get in Exult. They have this 'toggle footpads' mode which shows the entire boundaries in the Z-dimension of sprites.

I use it tons because it helps me see exactly where sprites end and begin, if there's 'empty space' surrounding a sprite that I can't see, it helps a lot. i.e. if the sprite is larger than the actual art shows. Having that in Exult would be helpful given the amount of construction I do and how many new sprites I'm adding

I see @wench has a version of this purely for the Shape Editor as a current PR. I guess I'm wondering if we could implement it as a map mode for all of the shapes all at once, like Pentagram has.

Pentagram screenshot:

Image

wench:
Just a small PR that adds showing the 3d bounding box to the ShapeInfo editor. The 3d Bounding box will only be shown when any of the 3 tiles spinner controls are changed

Showing the bbox by default can be done by inverting the `#if` at shapeedit.cc:2577

I added an int constant outline_color to Shape_draw that is the palette index used to draw the bbox. I also updated things so this constant is also used to draw the outline of selected shapes in the browsers. This is set to index 249 bright yellow

This is what it looks like
shape info

This  bbox drawing is slightly different to what I had already added to the ingame shape browser as it draws the rear lines behind the shape instead of drawing them all on top of the shape. I updated the ingame shape browser to use this new drawing method

If the sizeof the bbox exceeds the drawable area the drawable area and window will be resized bigger to fit 

Also added ability to show bounding boxes in Exult as requested by @thecommissar
 in #830. Pressing Ctrl-Alt-B will toggle showing bounding boxes and cycle through a selection of different colours. When there is a lot of clutter on tables  its difficult to see what's going on, but i'm implementing the feature  as requested
exult003 closes #830

### DIFF
--- a/mapedit/exult_studio.glade
+++ b/mapedit/exult_studio.glade
@@ -11151,7 +11151,6 @@
                                 <property name="adjustment">adjustment96</property>
                                 <property name="climb-rate">1</property>
                                 <property name ="value">0</property>
-                                <signal name="changed" handler="on_shinfo_shape_tiles_changed" swapped="no"/>
                               </object>
                               <packing>
                                 <property name="position">0</property>
@@ -11175,7 +11174,6 @@
                                 <property name="adjustment">adjustment95</property>
                                 <property name="climb-rate">1</property>
                                 <property name ="value">0</property>
-                                <signal name="changed" handler="on_shinfo_shape_tiles_changed" swapped="no"/>
                               </object>
                               <packing>
                                 <property name="position">1</property>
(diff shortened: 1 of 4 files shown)


wench:
@Dragon-Baroque  I've made changes let me know if you want anything else changed

PR #834, opened 2025-12-17
Exult Studio audio tester

DominusExult:
Added an Audio Tester (menu->tools) to play back music, sfx, speech from Exult Studio in Exult.
Also added play/stop buttons in various windows that allow setting a track (shape and egg settings).

When audio/music/sfx/speech is disabled in Exult, Exult Studio will display an alert.

image

### DIFF
--- a/mapedit/studio.cc
+++ b/mapedit/studio.cc
@@ -3043,6 +3043,19 @@ void ExultStudio::read_from_server() {
 	case Exult_server::select_status:
 		set_edit_menu(data[0] != 0, data[1] != 0);
 		break;
+	case Exult_server::play_audio: {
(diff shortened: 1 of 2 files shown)


wench:
Just a comment you'd probably want to override the audio being paused when Exult loses focus or at least unpause  the audio as if Exult regained focus when the sound test command is received by exult

wench:
I checked how the code works and  Audio::pause_audio only pauses currently playing audio samples it doesn't pause Audio generation entirely and doesn't pause midi playback so unpausing Audio here is actually entirely unnecessary with the way things are currently coded. Not pausing Midi is probably a bug and something I should fix

PR #852, opened 2026-01-01
New Usecode Intrinsic create_new_egg

DominusExult:
This allows creating all types of eggs with all the possible settings.

Also added optional frame setting to UI_create_new_object2.

exult_intrinsics.txt is a placeholder change for when it gets accepted to show wall the options before I add it to seventowers and let php create the file properly.

This was years ago discussed in https://sourceforge.net/p/exult/feature-requests/190/ by @KnightCaptainU7 and @drcode1 and I wanted to take a look. I then realized that creating an egg takes more than just creating the shape and frame to begin with.
@marzojr please take a look, especially if this is even really wanted.

### DIFF
--- a/usecode/intrinsics.cc
+++ b/usecode/intrinsics.cc
@@ -594,31 +594,34 @@ USECODE_INTRINSIC(create_new_egg) {
 	//   button(10):  [distance]
 	//   intermap(11): [dest_x, dest_y, dest_z, mapnum]
 
+	if (num_parms != 7) {
+		cerr << "create_new_egg: Expected 7 arguments, got " << num_parms
+			 << endl;
+		return Usecode_value(static_cast<Game_object*>(nullptr));
+	}
+
 	const int type = parms[0].get_int_value();
 	if (type < 1 || type > 11) {
 		cerr << "create_new_egg: Invalid egg type " << type << endl;
 		return Usecode_value(static_cast<Game_object*>(nullptr));
 	}
 
-	// Optional parameters with defaults
-	const int criteria    = (num_parms >= 3) ? parms[2].get_int_value() : 0;
-	const int probability = (num_parms >= 4) ? parms[3].get_int_value() : 100;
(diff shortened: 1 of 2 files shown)


usecode/intrinsics.cc
```diff
@@ -554,14 +554,215 @@ USECODE_INTRINSIC(create_new_object) {
...
+		if (frame != 0) {
+			obj->set_frame(frame);
+		}
+		UI_update_last_created(1, &parms[1]);
+	}
+	Usecode_value u(obj);
+	return u;
+}
+
+USECODE_INTRINSIC(create_new_egg) {
```
marzojr: Can you make this function take an array of data parameters instead of optional parameters that may or may not be used depending on type?

DominusExult:
I've fixed both. The frame shape was the usual oversight 🤦‍♂️ (especially annoying as I was wondering why the egg types enumeration doesn't begin with 0).

Array has been done. Tomorrow I'll write the proper documentation for SevenTowers and will update it properly before merging.

One style question: should I delete the lengthy instructions comments in intrinsics.cc once the docs are done properly?

usecode/intrinsics.cc
```diff
@@ -554,14 +554,223 @@ USECODE_INTRINSIC(create_new_object) {
...
+	}
+
+	// Optional parameters with defaults
+	const int criteria    = (num_parms >= 3) ? parms[2].get_int_value() : 0;
+	const int probability = (num_parms >= 4) ? parms[3].get_int_value() : 100;
+	const int distance    = (num_parms >= 5) ? parms[4].get_int_value() : 0;
+	const int flags       = (num_parms >= 6) ? parms[5].get_int_value() : 0;
+
+	// Extract data array elements
+	int user_data[5] = {0, 0, 0, 0, 0};
```
marzojr: This can be
```c++
int user_data[5]{};
```

marzojr (changes requested):
Just a few minor changes, but looks good overall.

PR #899, opened 2026-02-03
Added more user facing strings to exultmsg.txt

### LINKED ISSUE #895: Localize Oink!
With the ability to localize Exult I found some hardcoded strings, that I forgot again (will make a note next time).

But just now stumbled over `Oink! `in SI. It's hardcoded at several places in our code (as it replaces several texts).

gamewin.cc, CombatStats_gump.cc, Paperdoll_gump.cc, Stats_gump.cc, objs.cc, ucinternal.cc

So probably just needs two entries `Oink!` and `oink!` and point these at our textmsg.txt.

The gump files listed above, probably have more game specific hardcoded text that would need de-hardcoding.

DominusExult:
Fixes #895

@wench I began enumerating after the strings added in your savegame_changes branch.

When there are more than three messages in a file, I've made a Strings Class, except for exult.cc - IMO that's so big that I think it adds more clutter than commenting the string default in place.
But I can easily add a String Class to each file for keeping the pattern, hence leaving the PR for review.

### DIFF
--- a/cheat.cc
+++ b/cheat.cc
@@ -38,6 +38,7 @@
 #include "gamemap.h"
 #include "gamewin.h"
 #include "istring.h"
+#include "items.h"
 #include "miscinf.h"
 #include "mouse.h"
 #include "objiter.h"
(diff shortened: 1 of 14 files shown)


wench:
Looks fine. I created the classes to make it easier to use the translated strings and not to put too much unrelated stuff into items.h Just using constants from items.h for strings didn't seem like a good way to do things so I created classes with static methods that would do the string lookup correctly using the number from exultmsg and the offset.

The big thing still needing updating is the cheat screen and the way the key for the menu options integrate into the labels complicates things. Each option has a partial string for the labels and the key code needed to activate it. Would need to include the key codes in Exultmsg for it to be properly translatable but I don't really want to do that as the key codes are hard coded in switch statements and making them dynamic would be a fair amount of work. So making the cheat screen translatable is not something that I would do. Cheat screen layout is also dependant on fixed sized English labels. Order of options and label width would get all messed up with translations

DominusExult:
> I'll do the help screen after this is merged

Then I'll do the soundtester and shape browser (I guess you will do its help screen with the other strings in keys.cc).

wench:
0b7abeb has The Action description changes for the Help Screen
5a86c05bd625e4ddffcdb02dce8196ebf08ec708 has the shape browser help changes

PR #921, opened 2026-02-16
Update exultmsg_de.txt

ungesundes-halbwissen:
lines 1-700 translated
translation tidied up
search for "##" to find yet untranslated strings
Shape Browser and following remain untranslated as of yet.  Open question: Do we need to translate the editor part? Is working with Exult on this level possible, if you are not capable of speaking/understanding English (i.e. documentation etc.)

### DIFF
--- a/data/exultmsg_de.txt
+++ b/data/exultmsg_de.txt
@@ -283,9 +283,9 @@
 0x619:    digitale Musik
 0x61A:    Midi Treiber
 0x61B:SFX:
-0x61C:    pack ##
+0x61C:    Pack ##
 0x61D:Sprache:
-0x61E:    conversion ##
+0x61E:    Umwandlung ##
 
 #GameEngineOptions_gump.cc
 0x620:Normal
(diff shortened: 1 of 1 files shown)


DominusExult:
> Shape Browser and following remain untranslated as of yet. Open question: Do we need to translate the editor part? Is working with Exult on this level possible, if you are not capable of speaking/understanding English

Not really needed to translate these. Exult Studio is not yet (if ever) translated so the strings are only there for completeness sake. Same for the help screens (keys.cc).

Lines 152-154 are not correct yet.
```
0x57F:Du hast das Gleichgewicht
0x580:auf der Schlangeninsel
0x581: und |nur |genau | Jahr| Jahre| Monat| Monate| Tag| Tage| Stunde| Stunden|negative Zeit!
```
It's one sentence in which the last line can't have anything added (unfortunately, and I think the `&` needs to stay as well). so maybe:
```
0x57F:Du hast die Schlangeninsel
0x580:ins Gleichgewicht gebracht, in 
0x581: & |nur |genau | Jahr| Jahre| Monat| Monate| Tag| Tage| Stunde| Stunden|negativer Zeit!
```

`0x564:und mir Clerasil besorgen.` <- Not used the brand "Clearasil" on purpose?

```
0x5A2:Musik: 
0x5A3:MIDI Musik: 
0x5A4:OGG Musik: 
0x5A5:SFX: 
0x5A6:Sprache: 
```
On these you added space after the `:`, can you get rid of this?

PR #926, opened 2026-02-22
Add egg areas for all egg types

DominusExult:
With Alt-E you toggle Egg Areas.
It shows you a square around the egg when distance is a valid parameter of the egg.
Alt-E toggles the display for all eggs, every following Alt-E cycles through the egg types: monster, jukeboy, soundsfx, voice, usecode, missile, teleport, weather, button, intermap (and then turning it off again). An on screen message shows you the current egg type for the egg area display.

In Exult Studio's button bar you have a drop down list to set the egg type with egg area. And the egg editor will paint a poison green area around the egg being edited (which changes size on apply).

### DIFF
--- a/cheat.cc
+++ b/cheat.cc
@@ -47,6 +47,7 @@
 #include "touchui.h"
 
 #include <algorithm>
+#include <filesystem>
 #include <iomanip>
 #include <iostream>
 #include <map>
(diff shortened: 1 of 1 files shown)


wench:
The utility of this is so obvious I have no idea why I didn't have the code I added apply for all eggs. I must have at least thought of it or I wouldn't have committed the code that I originally wrote to test the jukebox egg behaviour to master

PR #928, opened 2026-02-25
Update exultmsg_de.txt

ungesundes-halbwissen:
Added translated egg area strings


DominusExult:
I've adapted your translation with my comments and committed that :)

Issue #731, opened 2025-06-09
SI Fixes - Requests collection


DominusExult:
- Harnna removing strange coin conversation point https://sourceforge.net/p/exult/feature-requests/206/

> When you talk to Harnna about the strange objects the question about the strange coins doesn't vanish from the conversation options when you clicked on it. 

- Give Neyobi a Schedule https://sourceforge.net/p/exult/feature-requests/175/

> Neyobi the Gwani child (NPC147) does not get out of Waiting when the Ice Dragon Blood is given to her mother, Yenani (NPC148), although the activity when she is double-clicked changes. Giving the blood Unsets NPC flag 30 on Neyobi, which is the SI_Zombie flag. If that flag is set, a companion will give a short speech saying that she is very sick. After the blood is given or the SI_Zombie flag is Unset manually, she will talk if double-clicked and walk to the Avatar, but after the conversation ends she walks back to her pelt and stands there in Waiting. Her schedule does not get updated to act like a normal NPC.
> In the DOS game the same behavior happens after the blood is given. She is available to talk but just stands there Waiting. Nothing happens if she is double-clicked before then.
> In neither case did I test after the Banes are released.
> I could not find a matching Global Flag in a quick scan of the list. If fixing this in SIfixes, it may make sense to give her the schedule to start with, but with the current business activity as Waiting. Then when the ice dragon blood is handed over switch her to Loiter or even have her hug Yenani or the Avatar.
> Steps to recreate:
> Start a new game.
> Turn on God Mode, and teleport to the Ice Dragon's inner lair. Kill an ice dragon, adult or baby, and fill the bucket that is conveniently there.
> Teleport to Neyobi's location, and check her behavior before and after giving the blood to Yenani who is usually in the same cave.

- Give Ice Dragons the Dragon Breath attack https://sourceforge.net/p/exult/feature-requests/174/

> In SI, the adult Ice Dragons (Shape 877) use the Telekinesis Shape 443 as their main attack instead of the Dragon Breath (Shape 857). Since the Telekinesis shape is made out of light whites and blues, it is barely visible in the snowy lands. This appears incorrect because the Ice Dragon shape includes smoke or steam in its attack frames.
> The Damage for on the Weapon tab for the adult Ice Dragon’s shape is only 4, so they do the same damage as their Baby Ice Dragons. This makes it much too easy to dispatch these dragons, considering in the story they killed three of the Gwani attack party.
[...]

PR #931, opened 2026-02-27
SIfixes Selina's Key Issue

KnightCaptainU7:
Adds source files to content/sifixes to address the concern from Crowley about the key that Selina starts with. To avoid binary map edits, this removes the key as part of the game start sequence, then adds to back to her in her final scene. This avoids the potential to skip the intended path to the antechamber. See https://github.com/exult/exult/issues/731 for details.


KnightCaptainU7:
For testing purposes:
Compile the Usecode, place it in the mod folder.
Start a new game.
In stdout.txt you should see SERPENT ISLE FIXES, which confirms the player is using a mod.
Check Selina NPC 44 to verify she does not have a key in her backpack.
Teleport to south of the Spinebreaker narrow hallway.
Walk north into the ambush antechamber, and defeat the 4 henchmen.
Selina's body should have the key on it. It is placed directly in her possession because it's possible to take her backpack away before she leaves the party.
The bodies of the 3 other henchmen should be identifed as "NPC's body".
Selina's body is not identified here in this test, she is MET in normal gameplay before this scene.

I did not add weapons and then warm clothing to the henchmen for now.

KnightCaptainU7:
image

KnightCaptainU7:
Correct. I ran into the issue with structs and could not compile and test. The code submitted works if there is a free spot or container. If merged in I will update it for your edge case.

PR #935, opened 2026-03-03
Improved giving Selina the key

KnightCaptainU7:
Improved the method used to place the key on Selina, in her backpack, and even if her hands are full and she's over her carrying weight limit. This should work in all scenarios.

### DIFF
--- a/content/sifixes/src/misc/egg_spinebreaker_antechamber_ambush.uc
+++ b/content/sifixes/src/misc/egg_spinebreaker_antechamber_ambush.uc
@@ -22,14 +22,49 @@
  *  Also sets MET on the other ambushers Palos, Deadeye, and Brunt.
  *
  *  2026-02-27 Knight Captain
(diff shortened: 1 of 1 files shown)


KnightCaptainU7:
See https://github.com/exult/exult/issues/731 for details. You should see this output in stdout.txt if she is wearing a backpack and they key is placed there.
```
Selina is wearing a backpack.
Selena's key created on her.
```

In the non-modded game, the key is in her backpack.

image

With this improved code, the key goes into her backpack if she is wearing it.

image

Even if she is over her carrying weigh limit, she still gets the key in her backpack. This is true with and without the hackmover still enabled.

image

Without her wearing a backpack it still works.
image

Stdout.txt will show she's not wearing the backpack but still gets the key.
```
Selena is NOT wearing a backpack.
Selena's key created on her.
```

I should probably turn these ~30 lines into a function, but for today, here we are.

PR #963, opened 2026-04-16
Actions updates for our Github workflows

DominusExult:
Updated the actions and outsourced the notify job to notify.yml as all CI workflows used the same script.
Changes to notify.yml will only trigger the ci-linux workflow to run to avoid wasting runner time.

### DIFF
--- a/.github/workflows/ci-android.yml
+++ b/.github/workflows/ci-android.yml
@@ -84,7 +84,7 @@ jobs:
           tls: true
           message: "\x0313exult\x03/\x0306${{ github.ref }}\x03 \x0314${{ github.sha }}\x03 https://github.com/${{ github.repository }}/actions/runs/${{ github.run_id }} The Android build \x034failed\x03."
       - name: Discord success notification
-        uses: rjstone/discord-webhook-notify@v1
+        uses: rjstone/discord-webhook-notify@v2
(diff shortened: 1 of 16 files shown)


DominusExult:
Instead of triggering the linux workflow on changes to notify.yml, the iOS workflow will be triggered as it is the fastest build.

DominusExult:
Fixing FreeBSD and OmniOS builds currently.

wench:
You fixed all the node warnings and fixed freebsd nice

DominusExult:
The warnings annoyed me and I was sure these would be easy to fix, however FreeBSD and OmniOS were a bit harder 🤦‍♂️🤩

The notify IRC and Discord actions still show the nodes warning.
If nothing happens until the deadline I'll have to find replacements.

wench:
The actions might just work when the upgrade is forced so even if the maintainers do nothing, our workflow might still work after the deadline. Still would be a good idea to make an issue if there isn't already one for both action. If everything goes wrong we now only need to fix it in one place. Could do as the warning message says and set the env to force Node 24 now to see if things will fail

DominusExult:
I did test them and they continued to work but then found replacements anyway - the Discord one had other visual issues.

Issue #976, opened 2026-05-28
Usecode should also display special characters


DominusExult:
Similar to how we allow displaying special characters for msgfiles we should also allow this for usecode.

So having something like äöü in usecode should make it find the correct font shape. Currently this will not display the character at all.

My proposal is to make the two character maps in files/msgfile.cc available for usecode as well.
Possibly dehardcode this and make it available in a file. This would allow an easy change for existing and no longer updated fan translations.

@marzojr when I take this on, I think I could use shape_info.txt for this as well with two new sections. Do you agree?

@ungesundes-halbwissen thank you for pointing me at this :)

DominusExult:
Decided to use a standalone file, font_map.txt, as it applies to any game not just either.

PR #978, opened 2026-05-28
De-hardcoded font map and apply it to Usecode

DominusExult:
The map for special characters in `./files/msgfile.txt` is now de-hardcoded to .`/data/fonts/font_map.txt`.
As per #976 Usecode output (conversations, books, signs) also needs the mapping which this also fixes.

The new source files for the mapping and de-hardcoding are ./shapes/font_map.cc/.h - I'm not entirely happy but I wasn't sure where else to place them and other fonts related sources are in ./shapes. If there is a better place, I can move them over.


wench (approved):
One small change I'd like to see a reverse method to get the UTf8 from the font char. Would probably output to a fixed size array for the utf8 chars.

DominusExult:
Something like this?

wench:
Yeah that's great

marzojr:
If I'm understanding this correctly, we're basically making usecode have UTF-8 strings which get translated at runtime to a font-based mapping? If so:
- we probably want to modify either `string_literal` or `Handle_string` in uclex.ll to validate UTF-8 and raise an error if it fails
- we may want to support some rune characters on the mapping files, in case people want to write text in runes (e.g., [this](https://github.com/marzojr/u7-usecode/blob/master/runes.txt))

These can be probably be done after.

marzojr (approved):

DominusExult:
Yes, you can already have UTF8 in usecode compiled by ucc, just Exult didn't know what to do with it 🤷‍♂️

Runic would need more thoughts to make it right and only having runes for signs so far.

PR #988, opened 2026-06-23
Combat start safely auto swaps in hand items with weapons now.

### LINKED ISSUE #987: Disabling Smooth Scrolling on slow walk
We could disable smooth scrolling for slow walking. This makes the ice skating go away, but otoh it clearly showcases what smooth scrolling does when it switches off and on.

I'll add a video to showcase this.

```diff
--- a/exult.cc
+++ b/exult.cc
@@ -1443,6 +1443,8 @@ static void Handle_events() {
 			Actor*        act   = gwin->get_camera_actor();
 			Barge_object* barge = gwin->get_moving_barge();
 
+			const bool slow_walk = !barge && act && act->get_frame_time() * 2 >= 3 * gwin->get_std_delay();
+
 			// Force a reset if position changed
 			if (last_x != gwin->get_scrolltx() || last_y != gwin->get_scrollty()) {
 				// printf ("%i: %i -> %i, %i -> %i\n", ticks, last_x,
@@ -1488,7 +1490,9 @@ static void Handle_events() {
 
 			// Is lerping (smooth scrolling) enabled
 			if (mswait && ticks < (last_repaint + mswait * 2)) {
-				gwin->paint_lerped(((ticks - last_repaint) * 0x10000) / mswait);
+				// Slow walk doesn't use smooth scrolling.
+				const int factor = slow_walk ? 0x10000 : ((ticks - last_repaint) * 0x10000) / mswait;
+				gwin->paint_lerped(factor);
 				didlerp = true;
 			}
 		}
```

_Originally posted by @DominusExult in https://github.com/exult/exult/issues/985#issuecomment-4759735234_

DominusExult:
Fixes #987
In the rare case that an item in hand can't go into another ready spot or a container, we "displaced" the item. It was sharing the ready spot with the swapped in weapon but was no longer visible. Now we drop the item to the floor (like the original SI) but we will never drop the spellbook to the floor, even if it is unusable at the moment. In this case the weapon won't be swapped in.

This bug also had potential to crash Exult AFAICT during my tests.

Alternative:
We could also never swap a weapon if an item would need to be dropped, I think I prefer the dropping.

### DIFF
--- a/actors.cc
+++ b/actors.cc
@@ -489,6 +489,10 @@ bool Actor::ready_best_shield() {
 	if (two_handed) {
 		return false;
 	}
+	// Don't replace a spellbook if it can't be stored in a container.
+	if (spots[rhand] && is_unstorable_spellbook(spots[rhand])) {
+		return false;
+	}
 	if (spots[rhand]) {
 		const Shape_info& inf = spots[rhand]->get_info();
 		if (is_in_party() || inf.get_armor() || inf.get_armor_immunity()) {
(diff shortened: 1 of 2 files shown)


i30817:
This is a easy way to lose plot critical items like the ophidian items thou.

DominusExult:
Yes, currently it was highly critical as the items just disappeared

PR #989, opened 2026-07-03
10 Enhancements

Qshadow:


wench:
This PR is doing too much with no description for me to even consider approving it.

objs/contain.cc
```diff
@@ -171,9 +179,84 @@ bool Container_game_object::add(
...
+	}
 	return true;
 }
 
+/*
+ *  Combine gold coins in this container into piles of 100.
+ */
+
+void Container_game_object::consolidate_gold(int shapenum) {
+	static int consolidating = 0;
```
rofl0r, line 193: this static int looks funky. under which circumstances can it be set when entering the function? if the function can be called by multiple threads (which seems unlikely), one would need a proper atomic test, else it's useless.

wench: I suspect this was AI generated. 2 commits indicate they were created with Claude so I suspect the rest might have been too. So many commits with commit times so close is also suspicious.

Qshadow: its AI generated, but it is tested and it works, I am half game past with them.

the large number of commits in one time is just because i collected them all locally during past month and pushed just today.

Qshadow:
> This PR is doing too much with no description for me to even consider approving it.

Yes, I will break them into separate, so you could check them one by one.
closing this one

PR #990, opened 2026-07-03
Auto-launch configured default mod

Qshadow:
- Add custom config/gameplay/default_mod key with an empty value when missing.

- If no game is selected by command line, use config/gameplay/default_mod to auto-launch the first configured game that contains a compatible mod with that exact name.

- If a game is selected by command line without --mod, launch that game's configured default mod when present and compatible; otherwise launch the selected base game.

### DIFF
--- a/exult.cc
+++ b/exult.cc
@@ -550,7 +550,9 @@ int exult_main(const char* runpath) {
 		}
 		config->remove("config/gameplay/allow_double_right_move", false);
 	}
-
+	if (!config->key_exists("config/gameplay/default_mod")) {
+		config->set("config/gameplay/default_mod", "", true);
+	}
 	// Setup virtual directories
 	string data_path;
 	config->value("config/disk/data_path", data_path, EXULT_DATADIR);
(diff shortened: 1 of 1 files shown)


DominusExult:
Honestly I don't favor that.
There are two reasons this is not a good idea:
- Exult Studio uses the same command line switches, you might need to make sure that this still works with your changes.
- For developing I (and I guess other devs) use the command line switches a lot to run the games directly. Having a default_mod interfere would be a huge annoyance.

And besides that you can already run a mod via the command line. I don't see much use for a default mod when you run via command line when you can just add the correct switch there.

Qshadow:
For developers command line is convenient, but not for simple gamers, they don't use command line, they don't even know about it (they are not devs).

I understand your concern too, the only proper solution is to just create a launch shortcut with default mode for users when installing each mod: eg:
Exult-Blackring.exe - Shortcut
and inside
_"Exult.exe --game blackgate --mod Keyring"_
but how can we let user know that he needs to drop this shortcut on his windows desktop?

 anyway, i am ok with dropping this commit if u feel this is not a good idea

DominusExult:
Normal users should just use our launcher. When they know how to edit the cfg, then they should be able to figure out the documentation of the command lines and how to prepare a shortcut.

PR #991, opened 2026-07-03
Add opaque shortcut bar background

Qshadow:
Paint the 320-pixel shortcut bar with the neutral grey palette index used by Trinsic cobblestone, then add contrasting top and bottom borders. This prevents game and menu artwork from reducing shortcut icon readability.


DominusExult:
image
I really don't like that. it reduces visible screen estate and IMO looks out of place in the game. Also makes the "transparent" setting of the shortcut bar obsolete.

DominusExult:
Just maybe as a third `Use shortcut bar` option but IMO it really needs to establish a different look for this.

PR #992, opened 2026-07-03
Auto-consolidate gold coins in party backpacks

Qshadow:
Merge scattered gold coin piles in each party member's backpack into
stacks of 100 plus a remainder pile. Consolidation runs when gold is
added or when the player opens a backpack or inventory gump.

Reuse existing pile objects in place instead of destroying and
recreating every stack, so gump positions and object identity are
preserved. Do not hook modify_quantity(), so spending or removing
gold is not undone by re-canonicalization.

Add c_gold_coin_shapenum (644) for the gold coin shape used in BG/SI.

### DIFF
--- a/actors.cc
+++ b/actors.cc
@@ -2441,7 +2441,37 @@ void Actor::update_from_studio(unsigned char* data, int datalen) {
 #endif
 }
 
+void Actor::consolidate_backpack_gold() {
+	Game_object* bp = get_readied(backpack);
+	if (!bp) {
+		return;
+	}
+	Container_game_object* cont = bp->as_container();
+	if (cont) {
+		cont->consolidate_gold(c_gold_coin_shapenum);
+	}
+}
+
+void Maybe_consolidate_backpack_gold(Container_game_object* cont) {
+	if (!cont) {
+		return;
+	}
+	Game_object* owner = cont->get_owner();
+	if (!owner) {
+		return;
+	}
+	Actor* act = owner->as_actor();
+	if (!act || !act->is_in_party()) {
+		return;
+	}
+	if (act->get_readied(backpack) != cont) {
+		return;
+	}
+	cont->consolidate_gold(c_gold_coin_shapenum);
+}
+
 void Actor::show_inventory() {
+	consolidate_backpack_gold();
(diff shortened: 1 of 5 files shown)


DominusExult:
I like the idea but I think a broader approach and a less hardcoded one would be better:
Add a section to shape_info.txt (maybe) called inventory_consolidation.
In that section add shapes (and as an option also frame) that should be auto consolidated. Coins (gold, filari, monetari...), lockpicks, bolts, arrows (both have several kinds), reagents.
And have the consolidation use thst info.

DominusExult:
And without looking at the code, how do you handle inventory space, encumbrance and which partymember to use for consolidation?
Players might use party members as explicit mules for this and that and not be too happy when everything gets redumped into the Avatar's backpack each time you open the inventory.
This might be a reason to reject this.
Or done for each party member individually?

Qshadow:
its done per backpack , all the gold in it - consolidates.

i30817:
Good idea. This might make my only real contribution to exult obsolete ( the switch to merge stacks automatically without the slider popup unless you press shift and vice versa if off). I wouldn't mind it be removed if this is added, it would then only be useful for external stacking. I guess moving from inventory too, causes that annoying popup, so it can stay.

I have some concerns with this on so called "non party member" backpacks/sacks/satchels because they may be intended by map designers to obfuscate something or seem natural (like spider web is a reagent and if you open a backpack with 30 individual webs it *seems* different and tells a different story than a stack - abandoned sack colonized by spiders vs alchemist bag dropped by accident).

So I'd request this only happened with active sacks (sacks in a party member inventory or inside another container in the inventory of a party member) 
Edit: checked the code and it seems you already check for party members. And possibly that "sack within sack" case depending on how "owner" works. Objection withdrawn.

i30817:
I dont recall how I checked which objects stack or if I even did (I think I just put a if else around the code that opens the slider), but I suspect you dont need to add any data to check which objects stack generally, because the game already needs that information when two objects of the same type are dropped in the same place top open the merge popup to make the distinction.

Personally I'd do it on "drop while open" too, not just on "open", even with the small risk of teleporting coins, if the stack its merged to is moved under the cursor it won't look horrible.

PR #993, opened 2026-07-03
Render selected VGA fonts in fixed colors

Qshadow:
Add optional monochrome remapping to the Font renderer and apply it to fonts 3, 4, and 6 using dark gray, black, and brown palette matches. Also force the tiny black UI font to black in both Black Gate and Serpent Isle so anti-colored source pixels cannot reduce text readability.


wench:
Fix build issues and post before and after screenshots showing what this PR actually does

DominusExult:
The tiny black font is done like this on purpose to make up of the shortcomings of the too few pixels. Changing this doesn't make it better IMO but worse, just look at `a`.
Original vs patch picture
image

DominusExult:
Your other change loses the 3D effect of the fonts (before/after):
image
image

wench:
I was expecting there may be some Problems with making fonts monochrome that were not designed that way by the original artists hence why I asked for  screenshots, but I wasn't expecting things to be that bad. The 3d sign fonts are ruined. Removing the antialiasing on tiny black will likely have a visual improvement when it is painted on a background that is not close to the original book or scroll gumps but I can't think of anywhere we do that

PR #994, opened 2026-07-03
Add 15 and 20 FPS gameplay choices

Qshadow:
Expose 15 FPS and 20 FPS in the Gameplay Options frame-rate selector. These custom higher game-speed values can now be selected and preserved through the normal options screen instead of requiring manual exult.cfg editing.


DominusExult:
I don't mind this addition at all :)

PR #995, opened 2026-07-03
Smooth actor movement between tile steps

Qshadow:
Track each actor's previous tile and movement timing, then offset body and equipped-weapon painting between adjacent gameplay positions. Expand dirty regions to cover interpolated locations so companions and NPCs move smoothly without changing pathfinding, collision, or simulation timing.

Use the existing Smooth Scrolling percentage as the interpolation duration and disable the feature when Smooth Scrolling is off. Skip teleports and long-distance moves, handle map wrapping and lift changes, and suppress actor offsets while the existing camera interpolation is already positioning the camera actor to avoid double movement.

Keep walking-pose animation at the original 2 FPS cadence even when gameplay FPS is raised. Store timing per actor and apply it consistently to Avatar path walking, party followers, ordinary path-walking NPCs, and schedule-driven NPC movement, while position updates and interpolation continue at the selected gameplay rate.

Add a Game Display Options toggle so the interpolation can be turned off independently of Smooth Scrolling.

### DIFF
--- a/actions.cc
+++ b/actions.cc
@@ -291,15 +291,11 @@ int Path_walking_actor_action::handle_event(Actor* actor) {
 	if (done) {    // In case we're deleted.
 		reached_end = true;
 	}
-	const Tile_coord cur        = actor->get_tile();
-	const int        newdir     = static_cast<int>(Get_direction4(cur.ty - tile.ty, tile.tx - cur.tx));
-	Frames_sequence* frames     = actor->get_frames(newdir);
-	int&             step_index = actor->get_step_index();
-	if (!step_index) {    // First time?  Init.
-		step_index = frames->find_unrotated(actor->get_framenum());
-	}
-	// Get next (updates step_index).
-	const int frame     = frames->get_next(step_index);
+	const Tile_coord cur    = actor->get_tile();
+	const int        newdir = static_cast<int>(
+            Get_direction4(cur.ty - tile.ty, tile.tx - cur.tx));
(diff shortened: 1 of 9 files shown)


DominusExult:
This looks horrible on my machine. I've spent some time two weeks ago to eliminate the ice skating as much as possible and you bring it back, not to mention the screen shaking on slow walk (And I couldn't find the setting).
And it doesn't turn off when you turn off smooth scrolling.

https://github.com/user-attachments/assets/308cb417-e35f-4c30-9f93-43a0b938ab00

Qshadow:
Hmm, strange, i dont have this screen jumps at all, i do have ice skating though, but i like it much better than tearing or stattering frames of the actors. I will add menu option asap

DominusExult:
Hold your horses, you might need to work with our latest git and not 1.12.x.

Ice skating has received no love, so even with a setting this is not likely to get accepted. Definitely not with this jerking which is probably due to us disabling smooth scrolling on slow walk in recent git.

Qshadow:
I agree that with this jumping it's unplayable. but I took the latest main and tested on it, I will retest again.
I maybe know why you don't suffer from the non smooth movement as much as I do: I am playing on 15 or sometimes 20 fps, on large screen (point, x2), and their movement looks terrible.

Qshadow:
ok, i retested, indeed in the latest version this fix is not needed, they already perfectly smooth!

PR #996, opened 2026-07-03
Improve conversation layout and portrait rendering

Qshadow:
Place up to three speakers in fixed middle, lower, and upper screen thirds, with each portrait immediately left of its text region. Use opaque speech and Avatar-choice backgrounds for readability at any resolution.

Treat only the Guardian's known face shape as oversized so high-resolution replacement portraits retain their dimensions. Prevent imported and rendered portraits from using animated or transform palette entries 224-254, avoiding flickering and accidental transparency.

### DIFF
--- a/mapedit/shapelst.cc
+++ b/mapedit/shapelst.cc
@@ -961,12 +961,13 @@ gint Shape_chooser::check_editing_files() {
 
 static int Find_closest_color(
 		const unsigned char* pal,    // 3*255 bytes.
-		int r, int g, int b          // Color to match.
+		int r, int g, int b,         // Color to match.
+		int last                    // First palette index to exclude.
 ) {
 	int  best_index    = -1;
 	long best_distance = 0xfffffff;
 	// Be sure to search rotating colors too.
-	for (int i = 0; i < 0xff; i++) {
+	for (int i = 0; i < last; i++) {
 		// Get deltas.
 		const long dr = r - pal[3 * i];
 		const long dg = g - pal[3 * i + 1];
@@ -991,7 +992,8 @@ static void Convert_indexed_image(
 		int            count,         // # pixels.
 		unsigned char* oldpal,        // Palette pixels currently uses.
 		int            oldpalsize,    // Size of old palette.
-		unsigned char* newpal         // Palette (255 bytes) to convert to.
+		unsigned char* newpal,        // Palette (255 bytes) to convert to.
+		int            last           // First destination index to exclude.
 ) {
 	if (memcmp(oldpal, newpal, oldpalsize) == 0) {
 		return;    // Old palette matches new.
@@ -1006,7 +1008,7 @@ static void Convert_indexed_image(
 	for (i = 0; i < count; i++) {
 		const unsigned char pix = *pixels;
 		if (map[pix] == -1) {    // New one?
-			map[pix] = Find_closest_color(newpal, oldpal[3 * pix], oldpal[3 * pix + 1], oldpal[3 * pix + 2]);
+			map[pix] = Find_closest_color(newpal, oldpal[3 * pix], oldpal[3 * pix + 1], oldpal[3 * pix + 2], last);
 		}
 		Write1(pixels, map[pix]);
 	}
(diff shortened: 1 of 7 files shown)


DominusExult:
The idea of playing the faces better is good but it lacks lots of fine tuning. For example it uses less space than available.
image
And instead of taking options (transparent backgrounds) please add options (a couple of opaque backgrounds).

DominusExult:
> Prevent imported and rendered portraits from using animated or transform palette entries 224-254, avoiding flickering and accidental transparency.

Can you give an example of when this is happening?

DominusExult:
It would make more sense to have a fixed sized rectangle in the middle of the screen with sensible sizes.
At original size 320x200 up to 420x250 don't do it at all, afterwards set the rectangle at maybe 420x250, regardless of the higher res (going by the optional resolutions for mobile devices 420x250, 480x300).
Original at 320x200:
image

Your patched version at 320x200 (adding the additional problem when there are even more topics to choose from, possibly going off screen):
image

Qshadow:
Oh, yes, the idea is mainly not for these default portraits, its for my next "portraits mod"
I will add my portraits here (i worked on them really hard, pls dont judge too harsh ;)
AI Arcadion_150x125_indexed
AI Batlin_150x139_indexed
AI British_150x125_indexed
AI Constance_150x125_indexed
AI Elynor_150x125_indexed
AI Forsythe_150x156_indexed
AI Iolo_150x156_indexed
AI Jaana_150x156_indexed
AI Julia_150x125_indexed
AI Katrina_150x125_indexed
AI Laurianna_150x125_indexed
AI Mariah_150x156_indexed
AI Nastassia_150x125_indexed
AI Nystul_150x125_indexed
AI Ophelia_150x125_indexed
AI Pamela_150x125_indexed
AI Penumbra_150x94_indexed
AI Rowena_150x156_indexed
AI Sentri_150x125_indexed
AI Shamino_150x125_indexed
AI Spark_150x125_indexed

The characters are (in same order)
Arcadion_150x125_indexed.png
Batlin_150x139_indexed.png
British_150x125_indexed.png
Constance_150x125_indexed.png
Elynor_150x125_indexed.png
Forsythe_150x156_indexed.png
Iolo_150x156_indexed.png
Jaana_150x156_indexed.png
Julia_150x125_indexed.png
Katrina_150x125_indexed.png
Laurianna_150x125_indexed.png
Mariah_150x156_indexed.png
Nastassia_150x125_indexed.png
Nystul_150x125_indexed.png
Ophelia_150x125_indexed.png
Pamela_150x125_indexed.png
Penumbra_150x94_indexed.png
Rowena_150x156_indexed.png
Sentri_150x125_indexed.png
Shamino_150x125_indexed.png
Spark_150x125_indexed.png

Qshadow:
So, these images use 224-254 colors and they flicker very ugly, that why i had to fix that. also they are much bigger than standard ( can be properly played only on high resolutions, so Yes I agree that at smaller resolutions this should be auto turned off), that why also I had to move things around a bit.
And of course the text readability without background was pretty bad, so background was a must for me.

DominusExult:
So this is very much out of scope then and vital information you should have laid out with your PR.

To prevent flickering or unwanted transparency make sure you don't use the last 15 (or so) colors of the palette. These are used for color cycling and transparencies.
I suggest you use smaller portraits.

Qshadow:
I wanted to let users create or just import their own portraits, and currently they cant, (its not possible not to use last colors of palette).
Also I didn't think about it, but I see where the problem is: I am playing on a large resolution (2560×1440, point, x2), 
and its impossible to read the conversations: they span from top left to top right, and my neck started aching after 5min already. While you test it on original resolution and it looks bad.
same is with size of portraits, on large resolution they look great! much better than the small icon size original ones.
P.S. 
Ah, I see now that you took the centering idea, great.

Qshadow:
In any case , we can make this optional menu feature. If the user wants larger imported portraits or original ones

DominusExult:
You *can* use your own portraits and with the right software you can prevent using the last few of the colors. We will not add code that prevents using transparency or color cycling.
That stuff like the conversation faces shouldn't size down at higher resolutions is a long standing topic of discussion. Changing this is planned but it needs a lot of code changes.

When you submit pull requests it is important to test with several aspects of the game, foremost the original resolution.

PR #997, opened 2026-07-03
Combat Speed Menu option

Qshadow:
Add a Combat Speed selector to Game Engine Options alongside the existing gameplay Speed choice, defaulting to 10 FPS. Track the user's configured gameplay delay separately from the configured combat delay; when combat is toggled on, the active frame delay switches to the configured combat rate, and reverts to the configured gameplay speed when combat ends. Also restore the correct delay when loading a save game.

Implemented against the current Strings::/get_button_pos_for_label()/yForRow() layout system rather than the old fixed colx[]/rowy[] arrays, adding a new "Combat Speed:" string (0x635) to data/exultmsg.txt.

Incidentally fixes a compile-breaking bug already present in the "Smooth actor movement between tile steps" commit, which referenced configured_std_delay/configured_combat_std_delay in gamewin.cc without ever declaring them in gamewin.h.

### DIFF
--- a/data/exultmsg.txt
+++ b/data/exultmsg.txt
@@ -296,6 +296,7 @@
 0x632:Cheats:
 0x633:Feeding:
 0x634:Enhancements:
+0x635:Combat Speed:
 
 #InputOptions_gump.cc
 0x640:Single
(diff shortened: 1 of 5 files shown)


DominusExult:
There are problems with the original combat changing its speed is the less likely one to tweak, I think.

PR #1001, opened 2026-07-08
Cheat screen rewrite

wench:
Didn't plan on doing this, it just kind of happened

It is an almost complete rewrite of the Cheat Screen. Almost complete because small amounts of code were uesful and I didn't see a need to rewrite it and risk changing behaviour of things that already work as I think they should.

Anyway the aim of this was to replace the old undesigned unstructured strange complex immediate mode menuing system with a new intentionally designed object based menuing system. Unlike the old system where each menu was split across something like 5 confusingly named functions that used shared global state to select input modes and large switch statements to determine what to do with input, the new system has a single function for each menu that creates a menu object that contains a list of MenuItems. Each MenuItem is a HotSpot and MenuCommand object. The Hotspot defines the keycode, label and position of the Menu item. The new hotspots are a greatly expanded upon version of the old Hotspots used for Mouse and Touch input. Menu Commands Determine what should happen when the MenuItem is selected. Each Menu Command has a vector of InputHandler objects that are used to get input from the user before the Menu Command is actually activated. Menu Commands also have a simple event model using Lambdas as event handlers, so now everything needed to define hat a Menu Iten should do is kept in the same place with the use of Lambdas that actually do the thing the Menu item should do

InputHandlers provide basic validation of user input with the Integer InputHandler always does bounds checking, so now all integer inputs have bounds checking unless the bounds are set to INT_MIN/INT_MAX. Integer Input Handler will always allow hex input by using the 0x hex prefix. There is a Game object Input Handler that allows entering a NPC number or if you press P will enter pick mode to allow picking an object from the game world. Empty input will default to the grabbed NPC if there is one.  The NPC input handler is a subclass of Game Object that makes sure the object is a valid NPC. It is an error to enter pick mode and select something that is not a Actor.

I split the Cheat Screen code into 2 files. I moved all the Actor related code into a new file. Having a 1000 line file  and a 2000 line file seems better than having a single 3000 line file. 

A minor change in the code. What was called the NorrmalMenu, is now called the  RootMenu

The basic  logic to running menus is now in the CheatScreen::RunMenu method insatead of each menu needing  its own logic for running menus

There shouldn't be many obvious user facing changes but I'll detail the most obvious ones here. Because hex input is now always allowed having separate hexadecimal and decimal teleport options was no longer necessary. There is now a single Tile Coordinate option to go with the Geographic Coordinates option. The geographic option now has hotspots to select North or South /West or East. The teleport to map option will now only show if there are multiple maps. The NPC option will now report failure when trying to teleport to a NPC that isn't on the map instead of silently failing. The NPC menu now shows what map a NPC is on . NPC loctation will now only be shown if the NPC has a valid Location on a map. the Teleport to NPC menu item will also now only show if the NPC is at a valid location.

New Teleport Menu showing Not on Map Error
Screenshot 2026-07-09 000017

In the business menu the list of activities now displays slightly differently on mobile. By default it should be the same as old where it displays 3 columns with a row size of 7 causing the rows to have no gap. If there is enough horizontal space it'll draw 4 columns with a row size of 8 which makes it more readable IMO. The regular non mobile display uses 3 columns with row height of 8 which is the same as old

 Old Mobile display
Screenshot 2026-07-08 231302

New Mobile Display if the horizontal resolution is at least 420 pixels
Screenshot 2026-07-08 231407

In this screen it can be seen that the ESC to Cancel messages have changed. Instead of showing the message separately. The [ESC] Exit hotspot changes to [ESC] to Cancel when the user is expected to enter input. This change occurs everywhere not just in this situation

The Flag Modfier item in the main menu has been replaced with a Usecode Menu. Flag Modifier is now found in the Usecode Menu. The Usecode Menu also has "Reload Usecode" and "Call Usecode" items."Reload Usecode" will reload the usecode file from disk. Useful if a modder has edited Usecode and wants the changes to apply immediately without needing to restart Exult. The "Call Usecode" item is also intended for modders so the can test call an arbitrary Usecode function. This item requires input of an object to run the usecode on. it will then ask for the usecode function number to call,  the usecode number  of the selected object will be prefilled if the object has a usecode function. It will then ask for the event number to use. This item is obviously extremely dangerous because it allows running usecode on the wrong object. I found and fixed a crash in the usecode Conversation class when I did something bad with "call usecode" while testing

Usecode Menu when Call Usecode asks for an object
Screenshot 2026-07-08 234218

Making an error when entering input will no longer be treated as fatal errors completely cancelling the input, the current input phase will be cleared and the user can try again. Messages no longer need to be manually dismissed with a keypress they will automatically dismiss after 1 second. The version info display now shows the Menu title of the current game/mod instead of a hardcoded string for just unxepanded BG or SI. 

Updated Root menu
Screenshot 2026-07-08 235200

Pretty much all the cheat screen is now translatable. Added about 400 extra strings to Exultmsg.txt. I don't really expect anyone to try to translate it but it's now possible. Pretty much the only parts that aren't translatable are some of the version info strings and an error message that would only really be seen if something catastrophic went wrong

Most menu items take their Keycode from the first char of the Label string. Because all the label Strings are in exultnsg, translations changing the labels in exltmsg can change the keycodes used by the menu items. The meu items  that are an exception to this are the ones that have an arrow key as their keycode. Only characters that correspond to a valid SDL_Keycode will work (though uppercase characters will work not just lowercase) this pretty much means any printable ASCII charcacter can be used.

Many of the strings in exultmsg use leading and trailing spaces to provide some coarse positioning of elements

Pressing unknown keys will no longer print a "random" character, it will now display a 3 hex digit scan code or ???

I'm sure there is probably more that could be said but thats all i can think of right now

Here's a complex example of what a new menu item and Menu Command is like. This is Call Usecode in the Usecode Menu. Trying to do this in the old system is what made me give up and decide a rewrite was necessary.

```
	// Call Usecode
	// 3 input handlers
	// One for Target Object
	// One for Usecode function number
	// One for event id
	command = std::make_shared<MenuCommand>();
	command->inputs.push_back(std::make_shared<InputHandlers::GameObject>(false));
	command->inputs.push_back(std::make_shared<InputHandlers::Integer>(
			false, 0, gwin->get_usecode()->get_num_funcs() - 1, false, Strings::UsecodeMenu::Prompts::Enterusecodenumber,
			Strings::UsecodeMenu::Prompts::InvalidUsecodeNumber, std::nullopt, [](InputHandlers::Integer* self) -> bool {
				return gwin->get_usecode()->function_exists(self->value);
			}));
	command->inputs.push_back(std::make_shared<InputHandlers::Integer>(
			false, 0, int(Usecode_machine::num_events) - 1, false, Strings::UsecodeMenu::Prompts::EnterUsecodeEventid,
			Strings::UsecodeMenu::Prompts::InvalidUsecodeEventNumber));
	command->events.begin_phase = [=](MenuCommand* self) {
		// Get the object from input 0 and get its usecode function number and prefill input[1]
		if (self->phase == 1) {
			auto object_h = static_cast<InputHandlers::GameObject*>(self->inputs[0].get());
			auto ucnum_ih = static_cast<InputHandlers::Integer*>(self->inputs[1].get());

			int ucnum = object_h->object->get_usecode();
			if (gwin->get_usecode()->function_exists(ucnum)) {
				char buf[16];
				snprintf(buf, std::size(buf), "0x%x", ucnum);
				ucnum_ih->SetInput(buf);
			}
		}
	};
	command->events.Activate = [this](MenuCommand* self, SDL_Keycode) -> std::shared_ptr<MenuCommand> {
		auto target_input   = static_cast<InputHandlers::GameObject*>(self->inputs[0].get());
		auto ucnum_input    = static_cast<InputHandlers::Integer*>(self->inputs[1].get());
		auto eventnum_input = static_cast<InputHandlers::Integer*>(self->inputs[2].get());
		auto usecode        = gwin->get_usecode();
		// Always Reload usecode before calling so modders don't need to do it manually every time they want to test a change
		usecode->read_usecode();
		// Calling usecode might do things that use a different event loop so first wait for button ups
		cscreen->WaitButtonsUp(true);
		usecode->call_usecode(ucnum_input->value, target_input->object, Usecode_machine::Usecode_events(eventnum_input->value));
		// apply palette because usecode may have changed it
		pal.apply(false);

		return {};
	};
	label = Strings::UsecodeMenu::MenuItems::CallUsecode;
	items.emplace_front(Hotspot(offsetx, maxy - offsety1 - 99, label[0], label + 1), command);

```


DominusExult:
Looking forward to this :)

DominusExult:
One other improvement tied to my layers branch:
To prevent the cheat screen getting smaller with higher game area resolutions we could turn it into a layer with a fixed size.
Same could/should probably done for the Exult start menu, the game menus, start a new game and the intros/endgames.

wench:
Yeah probably a good idea,

Other than the Activity display for mobile I talk about above nothing i did in the cheatscreen dynamically changes based on resolution I did try positioning some elements on the right edge of the screen but in higher resolutions  it just left a huge unwanted gap so i made sure the elements positioned on the right edge assume a max width of 320

DominusExult:
Ok, then that is something to do after this and my layers branch branch are merged.
Don't want to get into merging hell ;)

theGreyWanderer-uc:
Wow hell of an update, epic!

PR #1004, opened 2026-07-12
Smooth actor movement between tile steps

Qshadow:
Track each actor's previous tile and movement timing, then offset body and equipped-weapon painting between adjacent gameplay positions. Expand dirty regions to cover interpolated locations so companions and NPCs move smoothly without changing pathfinding, collision, or simulation timing.

Use the existing Smooth Scrolling percentage as the interpolation duration and disable the feature when Smooth Scrolling is off. Skip teleports and long-distance moves, handle map wrapping and lift changes, and suppress actor offsets while the existing camera interpolation is already positioning the camera actor to avoid double movement.

Keep walking-pose animation at the original 2 FPS cadence even when gameplay FPS is raised. Store timing per actor and apply it consistently to Avatar path walking, party followers, ordinary path-walking NPCs, and schedule-driven NPC movement, while position updates and interpolation continue at the selected gameplay rate.

Add a Game Display Options toggle so the interpolation can be turned off independently of Smooth Scrolling.

### DIFF
--- a/actions.cc
+++ b/actions.cc
@@ -291,15 +291,11 @@ int Path_walking_actor_action::handle_event(Actor* actor) {
 	if (done) {    // In case we're deleted.
 		reached_end = true;
 	}
-	const Tile_coord cur        = actor->get_tile();
-	const int        newdir     = static_cast<int>(Get_direction4(cur.ty - tile.ty, tile.tx - cur.tx));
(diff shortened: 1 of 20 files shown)


DominusExult:
Please make standalone pull requests against current git. No chain of committs that do different things not related to your PR.

PR #1009, opened 2026-07-13
SI defaultkeys: Fix comments/help text for walking vs. running.

mbanck:
Before, both keypad and shift-keypad had "run" comments which carry-over to the keybindings help screen. However, shift-keypad is walking, so fix the comments for those actions.

Before:
Screenshot from 2026-07-13 18-37-28

After:
Screenshot from 2026-07-13 18-38-11

### DIFF
--- a/data/si/defaultkeys.txt
+++ b/data/si/defaultkeys.txt
@@ -18,19 +18,19 @@ KP1		walk_south_west	# Run south west
 KP2		walk_south	# Run south
 KP3		walk_south_east	# Run south-east
 
-Shift-right	walk_east	1	# Run east
-Shift-left	walk_west	1	# Run west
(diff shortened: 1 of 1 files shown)


mbanck:
I don't have BG game data around right now, so could not tests that. It is missing the "Run" comments for shift-keypad, so maybe it works out-of-the-box there?

```
KP3             walk_south_east # Run south-east

Shift-right     walk_east       1
Shift-left      walk_west       1
Shift-up        walk_north      1
Shift-down      walk_south      1

Shift-KP7       walk_north_west 1
```

mbanck:
Yeah, Black Gate is fine, I tested it now.

DominusExult:
Thank you!

PR #1012, opened 2026-07-19
Fix display size going wrong on macOS

DominusExult:
Opening the Video options uses the configured settings, not the current window/display dimensions.

On fullscreen macOS with HiDPi these dimensions can be native max or 2x the value, possibly ending up with completely wrong dimensions. A running Exult can compensate for this but when you restart Exult this can result in crashes.

@wench could you take a look whether this makes sense for all platforms (Windows mostly :))? I think using the configured one instead of the actual one does make sense for all, but I'd rather hear your opinion.

I had mentioned this problem somewhere in our issues before but it was kind of not always reproducing but now it was.

### DIFF
--- a/data/exultmsg.txt
+++ b/data/exultmsg.txt
@@ -390,6 +390,8 @@
 0x695:AR Correction:
 0x696:Same settings for window
 0x697:and fullscreen:
+0x698:Current Mode:
+0x699:Current Size:
 
 #imagewin.cc scaler displaynames 
 0x69A:Point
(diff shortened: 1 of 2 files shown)


wench:
I can't think of a reason why this wouldn't be appropriate for all platforms.

It may be a good idea to have a line of text in the gump that shows the current resolution

DominusExult:
Hmm, yes, I'll see if I can squeeze something in.

DominusExult:
@wench 
it would look like in the screen shot below, but this line needs to be ifdef'ed out for macOS, iOS and possibly Android (need to test) as it shows the same wrong value as it currently does for Display Mode :)

image

wench (approved):

PR #1017, opened 2026-07-26
Natural light for light sources

DominusExult:
Instead of just changing the game palette when light sources are lit or you get near them, every light source gets a layer with its light in a radius around it. This brings a whole new atmosphere to Ultima VII, just playing around with it I never want to use another way :)
Can be toggled in the Game Display Options or in the exult.cfg in config/gameplay/natural_light.

The way it works is that a light source gets its own copy of the game world with the light's palette applied while the game world remains at its daytime palette. This is likely a huge performance hog, especially in areas with many light sources, but I have not made any performance measurement.

Buildings have two new sections in the respective shape_info.txt files:
light_passes_through: 
these shapes or just frames allow light to pass through (open doors, open shutters, glass windows, broken walls, barred prison doors,... there is a lot). Light does not come out of a building if the light is blocked in, if there is an opening or a non-blocking shape, light will pass out. But indiscriminating, everything in the light radius will be lit up (or pass into other buildings that are not blocked in)

roof_shapes:
Roof shapes are excempt from being lit up from the inside, by adding a mask for roof shapes. Outside light sources will light the roofs in their radius.

Small and medium issues:
- broken roof shapes count as non-blocking but will only light up their shape, not more of the roof, probably could be made into a radius around it as well.
- light circles in SI are used to illuminate things secretly (e.g. parts of the dark path and their serpent gates) and they give off a glow now. Could be solved along with the next issue.
- Second floors, or roofs made out of floor pieces. This is tricky, along with declaring floor pieces at z >= 5 as roofs one would need to add more pieces to the no-light mask (for example the castle battlement) to prevent the roof being unlit but the shapes on the floor being lit.
- radius and light strength needs to be adjusted. The candle light is probably too low.
- lava in SI needs work as it fades out too quickly, gives off not enough light.
- more light sources: fireball traps, explosions, spells (fireworks, explosion, lightning...) etc.
- add radius for light spells (glimmer, light, great light)
- maybe add radius for the Helm of Light, but it may just keep on changing the whole game world palette.
- sometimes after starting up a game there might be brief shake up when the light layer goes momentarily out of sync with the game world.

Big issues with probably high performance impact (and which I'm not sure how to take on):
- Have solid objects with z >= 4 (?) block the light, either adding a mask for everything behind it or have the layer created with these measurements.
- only let light out at openings/non-blocking shapes, again either by adding masks or having the layer created with these measurements.
- masks need to be handled carefully to prevent objects "over" them not to be masked as well.

Some early screenshots
image

image

image

and a link to a recent video of the code in action
https://youtu.be/OMIuciLWW00


DominusExult:
@wench @marzojr @Dragon-Baroque @drcode1 
I'd be grateful if you could give it a spin and tell me of issues other than the above.

DominusExult:
@wench @Dragon-Baroque do you want to take a look at it? I think it is mostly ready to land in the source, gamewin.cc, gamerend.cc need a lot of comments clean up, though.

Dragon-Baroque:
Natural light is very beautiful, and much more convincing that the original U7 lighting.

Natural light gives me a major problem of performance, though.
Exult at Game 320x200 in window 1024x768 is somewhat slow.
Game Auto in window 1600x1200 and Scaling x1 is very slow : 100% one core on four cores AMD 5700G by htop in Linux. I don't think Exult is able to use more that one core.
I mean : Mouse movements are late and not smooth. Avatar's movement from the keyboard is also very slow.

Then, I have another problem, present whether Natural Light is on or off, that the configuration gumps are lighted like the game area.
Which makes them impossible to read at night ( Trinsic NE corner at the beginning, switch to 22:00 with the cheats to test the Natural Light, open the gumps with Escape ).
Except for the Video Options gump which, while open, is fully lighted ( this  is good ), but also switches the game to full day light ( this is amusing ). Could all gumps act like the Video Options gump ?

DominusExult:
I feared the performance issues and I eliminated one that was obvious on my machine but otherwise wasn't able to experience. Hence it has to be an optional setting and not the default. Unless @wench wants me to shelf it for a better solution than to have the lighted area be an extra layered copy of the game world.

@Dragon-Baroque the palette for gumps can be adjusted in Exult since recently with my addition of layers for UI elements. In the UI Options you can choose a palette being used when the game palette is not the daylight one. It can even be adjusted for several UI elements groups if you want.
There is a bug when you start Exult that it can't tell the game palette but I that @wench's savegame branch fixes that (it fixes a similar bug), so I didn't bother with that.

DominusExult:
@Dragon-Baroque I did some optimizations, but these probably are not entirely helping.

rofl0r:
still waiting for some up-to-date screenshots :)

Dragon-Baroque:
> @Dragon-Baroque I did some optimizations, but these probably are not entirely helping.

_Way better_ at Game Auto at Scale x1 in Window 1600x1200 :
- CPU still one core 100%, but mouse much more reactive, and less jumpy.
- Avatar's movement still noticeably slower than without Natural Light, but much smoother.

Any further gain you could win would be welcome, but I think Natural Light _is now_ playable. Thank you.
Any chance that Exult in Game Auto at large resolutions, with or without Natural Light, might use several threads ?

To compare : _Without Natural Light_, at same resolution, CPU one core around 70%, mouse and Avatar's movement are smooth, no slowdown compared to Game 320x200 in 1024x768.

wench:
I'll check this out shortly after dinner. I'll add some metrics to see exactly what performance is like for me,

Dragon-Baroque:
> @Dragon-Baroque the palette for gumps can be adjusted in Exult since recently with my addition of layers for UI elements. ... There is a bug when you start Exult that it can't tell the game palette but I that @wench's savegame branch fixes that (it fixes a similar bug), so I didn't bother with that.

When I set the Night Palette to Day in the UI Options and Apply it, it takes effect right away.
But when I start Exult anew, it is lost and I have to open the UI Options where the Night Palette has kept Day and click Apply again to make it work.
Is this what you meant by "There is a bug..." ?

DominusExult:
@Dragon-Baroque that is exactly the bug. If you reload the savegame the palette is correctly applied. I'm pretty sure that is fixed in the savegame branch.

As for the CPU usage, I can try to squeeze out a bit more but it might not be possible without more threads and/or leveraging SDL to use the GPU. And that is another big change on top.

Dragon-Baroque:
> @Dragon-Baroque that is exactly the bug. If you reload the savegame the palette is correctly applied. I'm pretty sure that is fixed in the savegame branch.

Is the use of the Quick Save a different code path from a Savegame load ?

For I am now keeping the `~/.exult/` folder and `~/.exult.cfg` file during my builds - I was removing them previously, by fear of compatibility -, and now I have Quick Saved after the Avatar's arrival sequence in Trinsic in Forge of Virtue.
When I now start Exult, it is past the arrival sequence, thus the Quick Save has been loaded, but the UI Night Palette set to Day is not taken.

Question, by the way : What changes in Exult may make either `~/.exult.cfg` or `~/.exult/` invalid ?

wench:
Trying this out it looks really nice . Performance seems ok to me. Found a problem that probably isn't due to these changes. I changed my display settings and my mouse position became out of sync. I'll try replicating it in master and if I can I'll make an issue for that

Also had this happen after changing display settings:
Screenshot 2026-08-07 215928
The problem went away when I put my setting back to the way they were. Changing them again caused the problem to come back

I have my settings 1280x960 x3 fill and i was switching between auto and 320x200. Auto was how I started exult and problems occurred when switching to 320x200

The out of synced mouse clicks are also occuring at the same time. In the screenshot the visual glitch actually shows the offset of the mouse clicks, If I want to click on the throne I need to click on the throne in the glitch not where it actually is. Everything on screen seems to have the same offset

DominusExult:
I think the glitches might be due to the fork not being in sync with the fixes for fill mode, fill scaler and aspect correction in main. 
Why do you insist on using anything other than fill mode fill???? :)

wench:
Mouse click offset is not replicable on Master so it's a problem in this branch

wench:
Rebasing your branch onto master fixes the problems

There are no conflicts so you should rebase your branch and force push it.

DominusExult:
I've did some more performance improvement and rebased on master.

*However* there is now a bug with the SDLScaler and the 320x200 game area. Everything masked out is now pitch black. At *auto* it is rendered correctly. For me the SDLScaler only uses the metal backend, do you experience the same with your SDLScaler backend?
It does look interesting :) 
image

Dragon-Baroque:
I built your fork at :
```
* in_vs_out_light  14865c05d [origin/in_vs_out_light] Books on shelves on higher floors were not illuminated.
```
and I got that :

The Avatar is under the Gump.
If I change the Renderer to OpenGL instead of Vulkan, same result.
After I wrote the info stuff below, I checked all other renderers, same result.
If I switch `AR Correction` to `Disabled`, same result.
This is the darkest corner of Trinsic at Day 0, 22:00.


Dominus

I you build Studio with `--with-debug=messages`, you can get some info at start :
```
    Hint RENDER_DRIVER is 'vulkan', Current Render Driver is 'vulkan'
    [ 0 ] Video Renderer is 'opengl'
    [ 1 ] Video Renderer is 'opengles2'
    [ 2 ] Video Renderer is 'vulkan'
    [ 3 ] Video Renderer is 'gpu'
    [ 4 ] Video Renderer is 'software'
```
Then you can switch the renderer with, for example, `SDL_RENDERER_DRIVER=gpu <pathToExultStudio> <Options>`.

wench:
Playing around with performance metric at 1280x960 turning on natural lighting pretty much halved my performance

Here is a Natural lighting off shot
Screenshot 2026-08-10 020743
And one with it on
Screenshot 2026-08-10 020731

Frame time increased from 22161us to 39500us

Looking at the details breakdown almost all of that time was being used to composite the light layers. I'll look a bit more into why compositing seems a little slow the layer for my Metrics text is also kind of slow to composite, it may just be SDL being slow but I don't think that would be it

Dragon-Baroque:
> Then you can switch the renderer with, for example, `SDL_RENDERER_DRIVER=gpu <pathToExultStudio> <Options>`.

**SDL_RENDER_DRIVER=...**, sorry.

Question, Dominus, is the layer with the  Avatar and Iolo and the torch by any chance the top layer, with the buildings that should be a bit lighted on some lower layer ? Then it seems that transparency / compositing does not work well.

DominusExult:
@Dragon-Baroque the SDL renderer list comes from Exult's debug messages, not Studio.
For me all render everything but the light layers black. But only since I rebased the fork with master.
Something something wrong with whatever I broke with 695064fda107a921229958e990fe858d67853b13 and fb01b2902c4bbea87e2c76a14fd144d371630421

Dragon-Baroque:
> @Dragon-Baroque the SDL renderer list comes from Exult's debug messages, not Studio.

Way too hot at home, cannot think or write properly. You are right, this is Exult all over, not Studio.

> For me all render everything but the light layers black.

You are using the layer architecture to provide the Natural Lighting, aren't you ?
Then I wonder, you might have some transparency issue.

Say, if you remove the call to `composite_layers` in `UpdateRect`, what do you see ? Nothing or the content of the `surf` `SDL_Surface` ? In the second case, what do you see when you restore the call to `composite_layers` but alter that one to display all layers but the last one ?

DominusExult:
My SDLScaler issues were also caused by my 695064fda107a921229958e990fe858d67853b13 and fb01b2902c4bbea87e2c76a14fd144d371630421. Obviously I've made a mess with these.

So @wench and @Dragon-Baroque should I commit this PR now and then revert  695064fda107a921229958e990fe858d67853b13 and fb01b2902c4bbea87e2c76a14fd144d371630421?

wench:
I have a few scaler related changes that I was planning on making a PR for over the next few days that fix things. I would suggest holding off making any changes because they will surely conflict with my changes. If just reverting those 2 commits fixes things then do it as I started my fixes by doing that.

Dragon-Baroque:
Lack of time, I can no more that contribute the following, for common understanding, perhaps as a starting point for some documentation, later on.
So this is left between you two.

The Exult display pipeline is made of _three_ Scalers, not _two_, though at least one of those is Scaling trivial.
The pipeline performs two Conversions in addition to Scaling, from 8 bits pixels with Palette to 32 bits pixels ( between Fixed Scaler and Fill Scaler ), and from CPU RAM `SDL_Surface` to GPU VRAM `SDL_Texture` ( SDL Scaler ).

1. The Fixed Scaler in `Image_window::show` jumps from the `draw_surface` `SDL_Surface` to the `inter_surface` `SDL_Surface`, unless it is trivial and then `inter_surface` is set to `draw_surface`.
2. The Fill Scaler in `Image_window::show`, when not `SDL Scaler`, jumps from the `inter_surface` `SDL_Surface` to the `display_surface` `SDL_Surface`. When `SDL Scaler`, it is not done at all and the `inter_surface` is shipped to the SDL Scaler in `Image_window::UpdateRect`. _I wrote this, but this now looks like a mistake to me, it should set `display_surface` to `inter_surface` and let `UpdateRect` make appropriate decisions based on `display_surface == inter_surface`_.
3. The SDL Scaler in `Image_window::UpdateRect` _should_ jump from the `display_surface` `SDL_Surface` to the `screen_texture` `SDL_Texture`. It converts the `SDL_Surface` to an `SDL_Texture` always, it applies GPU Scaling when the source `SDL_Surface` ( _it should be `display_surface` but it is `inter_surface` then_ ), converted to an `SDL_Texture`, has a smaller size that the target `SDL_Texture`.

I know that the case `Fill Mode Center` is missing somehow in the above, and I sure would like to understand how the                            Layer Architecture fits into it.

PR #1023, opened 2026-07-29
Translucency, Palette Cycling and Shape Animation for Exult Studio

### LINKED ISSUE #1013: Exult Studio translucency rendering issue
Came across this by accident, while comparing something I was trying to fix elsewhere I noticed that Exult Studio will not paint translucency to shapes it displays. For example have a look at shape 177 in ES:
Image

Its showing the cycling orange stroke and not applying translucency to it. It should appear like:
Image

My issues are slightly different from what ES' will be, but I think ES can get away with a much easier update.

For ES I think what is required is to re-use things that can already do it?:
- give draw_shape a bool translucent = false (shapedraw header)
- give shape_shape_single a bool translucent = false (shapedraw header)
- add new explicit-window paint_shape overload on shape_manager (shapeid.h) existing one uses static scrwin, unset in Studio, unsafe
- have overload call shape_manager's new overload instead of shape->paint, passing iwin (shapedraw)
- compute translucent via shape_manager - pass it through (shapedraw)
- have shape_shape_single forward translucent in all calls to shape_draw (shapedraw)
- swap raw paint to draw_shape in shapelst (line **165** only) and npclst
- chunklst: skip draw_shape, call shape_manager new overload with own rwin

I think this is the short path, but may not be desired. The other option is to deploy code to the paint stuff to do it properly.

wench:
closes #1013
Only originally meant to implement Translucency but decided to implement palette cycling and then Animation too

Not really much to say about this but I will anyway

Palette cycling and Animation is controlled by a 100 ms g_timeout in the ExultStudio class

The ShapeDraw class now maintains an iterable linked list for every ShapeDraw object. I use this to update the palette and animation for Each ShapeDraw and queue a draw  or perform a render. I also use this to fix a problem where changing the background colour in preference would only affect the current chooser and would not affect any other shape draws that has been created like shape singles in the shape info dialog or  choosers that were viewed and are now inactive 

Shapes from shapes.vga will render translucent based on shape info, other shapes (eg faces and sprites) are generally always painted as translucent because it seems to be more correct. Shape info editor paints the current shape as transparent based on the checkbox state

All three new capabilities have settings to control them. They are all enabled by default. Disabling all three settings should make things act identically to how they were before these changes

Shape Info is now aways loaded.  This shouldn't really have any impact but I thought I should note it

Xform tables are now loaded if Translucency is enabled, When loading xforms I modify the tables so the value  for the background colour (index FF) in each table  is set to the value of the closest colour in the palette. This generally looks fine with the expected quality that xforms give for 8 bit palettized translucency. If the background colour is equally unlike every colour in the palette it may not look great but I have no idea what colour that would even be like. A basic grey background has no problem because there are lots of grey and near grey colours. The xform loading doesn't try to auto generate xform tables if they are missing like Exult does. If someone needs Exult Studio to generate xform tables, IMO, it should be a Tools menu option perhaps with a dialog that would allow setting the RGBA blend parameters to create custom tables

Shape animation is not entirely identical to how things are in exult. Hour based Frame animation is set to advance frame every 10 real seconds instead of the 1 in game hour exult users which would be about 2 and a half real minutes, It seemed a bit slow to replicate it exactly. Random pause on first frame of generic loop animations is not implemented. The way I'm doing things pretty much synchronizes all animations and random pause would require me to do things differently. While not perfect it should be good enough. Changing thins that could affect shape animation will reset the animation and pause it for 2 seconds. Setting a shape chooser to Frames mode disables animation for the chooser.

Shape Browsers will size animated shapes to the maximum size of all frames if animation is enabled.This prevents horrible overlapping of shapes that change size when animating

I added a nested class to ShapeSingle to automate connecting to the changed singals from widgets. To reduce the amoun of boiler plate code needed to do this. Connecting to the signal with my class is as simple as declaring a field in a ShapeSingle class like this:
`	WidgetChangedConnect shape_3d_x    = {"shinfo_xtiles", "changed", on_widget_changed, this};`  The constructor parameters are the widget's id, the signal to listen too, the call back function and the user_data. The object helpfully keeps a pointer to the  widget. Using the example above, you can get the widget with `shape_3d_x.widget` which replaces `shape_3d_x_widget` that we had before

I set the table in the preferences window to be row-homogeneous so my new rows ere the same size as the existing ones. My checkboxes are just so much smaller than the other controls

Video showing all these changes. Everything is disabled at the start showing how things are now and then I enable all settings and the differences can plainly be seen. The Energy field shape is animated and translucent. Enabling animation causes it to be sized to its maximum size and the rest of the row is affected by the change. With the settings off I have it set to frame 15 that causes it to overlap with other shapes

https://github.com/user-attachments/assets/b640cdbf-519e-482b-98c5-6f8f981fdde9

### DIFF
--- a/mapedit/shapedraw.cc
+++ b/mapedit/shapedraw.cc
@@ -861,7 +861,7 @@ Shape_single::WidgetChangedConnect::WidgetChangedConnect(
 }
 
 Shape_single::WidgetChangedConnect::~WidgetChangedConnect() {
-	if (connect) {
+	if (connect && g_signal_handler_is_connected(G_OBJECT(widget), connect)) {
 		g_signal_handler_disconnect(G_OBJECT(widget), connect);
 		connect = 0;
 	}
--- a/mapedit/shapedraw.cc
+++ b/mapedit/shapedraw.cc
@@ -227,6 +227,9 @@ Shape_draw::~Shape_draw() {
 // Get animated frame using an Animation_info object
 
 int Shape_draw::GetAnimInfoFrame(const Animation_info* aniinf, unsigned short first_frame, unsigned short shape_frames) {
+	if (!ExultStudio::IsShapeAnimationEnabled()) {
+		return first_frame;
+	}
 	//
 	// Variables that represent the fields in the Frame_animator class
 	//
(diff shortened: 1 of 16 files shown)


wench:
Figured out  a way to implement Random pause on First frame for shape Singles without needing many changes. Wont affect animations in a shape list but implementing it properly in the Info Editor will be useful for people editing/creating animated shapes

mapedit/shapedraw.cc
theGreyWanderer-uc, line 1: Cool

wench:
Its been almost a week if there are no objections I plan on merging this tomorrow

DominusExult:
Just took a look. Looks really nice.

I got this build warning:
```
.././../../exulting/exultwench/mapedit/chunklst.cc:252:17: warning: '&&' within '||' [-Wlogical-op-parentheses]
  252 |                                         if (shinfo && !shinfo->has_translucency() || xforms.empty()) {
      |                                             ~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ ~~
.././../../exulting/exultwench/mapedit/chunklst.cc:252:17: note: place parentheses around the '&&' expression to
      silence this warning
  252 |                                         if (shinfo && !shinfo->has_translucency() || xforms.empty()) {
      |                                                    ^
      |                                             (                                    )
.././../../exulting/exultwench/mapedit/chunklst.cc:263:17: warning: '&&' within '||' [-Wlogical-op-parentheses]
  263 |                                         if (shinfo && !shinfo->has_translucency() || xforms.empty()) {
      |                                             ~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ ~~
.././../../exulting/exultwench/mapedit/chunklst.cc:263:17: note: place parentheses around the '&&' expression to
      silence this warning
  263 |                                         if (shinfo && !shinfo->has_translucency() || xforms.empty()) {
      |                                                    ^
      |                                             (                                    )
2 warnings generated.
.././../../exulting/exultwench/mapedit/shapedraw.cc:241:17: warning: variable 'frame_counter' set but not used
      [-Wunused-but-set-variable]
  241 |         unsigned short frame_counter;    // When to increase frame.
      |                        ^
```

DominusExult:
No more warnings on macOS :)

Issue #1024, opened 2026-08-03
sfx lack mod support


theGreyWanderer-uc:
Mod support for music override and extensions exists, but does not for SFX.

This can be easily modeled after the music in Audio.cc file for sfx by using some shared concepts:
numeric ID
- format three-digit filename
- check patch folder: PATCH/sfx/
- use patch when present
- if not use existing source

I have a working version of this:
- example track 1 override
- example track 175 addition
https://github.com/user-attachments/assets/62c5267f-065b-45dc-b0b4-78fddea702e3

Added a few guards with it:
- no playing negative numbers
- file size validate before conversion
- fails for invalid/valid/missing patch

PR #1025, opened 2026-08-03
sound effects override and extension based on music override and exte…

theGreyWanderer-uc:
RE: https://github.com/exult/exult/issues/1024

### DIFF
--- a/audio/Audio.cc
+++ b/audio/Audio.cc
@@ -45,6 +45,7 @@
 #include <cstring>
 #include <fstream>
 #include <iostream>
+#include <limits>
 #include <set>
 
 // #include <crtdbg.h>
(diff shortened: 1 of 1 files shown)


DominusExult:
have you extensively tested that it only patches those sfx that are present as numbered wave files? But it looks good and makes sense

theGreyWanderer-uc:
> have you extensively tested that it only patches those sfx that are present as numbered wave files? But it looks good and makes sense

The wav files that do not occupy a current SFX #(like 175) do not appear to play where they aren't expected. 

When I intentionally want to "overwrite" a SFX by putting the wav file in with a SFX #(like 1) which already exists, it has the intended behavior of patching in and playing that mod #1 wav over the default #1 wav.

I haven't done more than 5 or 6 patches though if you want me to test with a higher number I can.

Does that cover your question?

theGreyWanderer-uc:
> have you extensively tested that it only patches those sfx that are present as numbered wave files? But it looks good and makes sense

FYI video in: https://github.com/exult/exult/issues/1024

PR #1026, opened 2026-08-05
Add Spectron sidecar and optional C++ event bridge for Exult

Dhghomon:
This is an integration to test SurrealDB's memory context layer [Spectron](https://surrealdb.com/spectron) that is currently in early access. Thanks to Exult I was able to pipe events to an egui app in Rust which sent events to Spectron and would query it from time to time during the game to provide relevant info and show what the Avatar might be thinking as you play the game.

First blog post on why Ultima VII is the best choice for this sort of thing: https://surrealdb.com/blog/using-ultima-vii-to-test-spectron-part-i-the-game

ultima_screenshot_5


Dhghomon:
Oops, this should have been to my fork. Closing now

PR #1029, opened 2026-08-15
Scaler, surface and layer changes

wench:
Some changes to how scaling and surface creation works.

Mostly this is to fix SDLScaler and to improve efficency by removing the copying between Display_surface and Screen_texture

Primarily this is accomplished by getting rid of display_surface in the Image_window class and instead directly drawing to screen_texture by using SDL_LockTextureToSurface to create a temporary display_surface in Image_window::show().
Image_windows::updateRect() no longer takes a surface as a parameter but now takes a SDL_FRect that contains the region of screen_texture that needs to be rendered.

As display_surface no longer exists, inter_surface can't be set to it when creating surfaces and instead will be set to nullptr to indicate that it doesn't exist. in show() inter_surface will be set to the temporary display_surface as needed. 
Screen_texture no longer has the same size as the displsy resolution. It now includes the guardband. This was done to simply things an so the guardband offset hack could be removed and to ensure that edge clamping works as expected when using SDLScaler. Texture edge clamping on non pow of 2 textures is dependant on Hardware support so it was just easier to assume it isn't supported and just use the guardband to ensure consistent behaviour. 

SDLscaler behavior has slightly changed. The SDL scale mode used when SDLScaler is in use will now match what  Scaler is set to. If Scaler is point and FillScaler is SDLSCaler,the SDScaler will use nearest scale mode and scale the same as point. The allows for complete hardware acceleration of point scaling . Does have the downside that its no longer possible to use Scaler Point and use SDL for HW accrlerated bilinear aspect correction. The visual difference is so small that I didn't think it really mattered if I made this change in order to get fully HW accelerated point scaling.

This set of changes includes my Performance metrics overlay,toggleable with Alt-2. The overlay show how long it takes to do scaling and how long certasin fperfrmabce critical bits of code are taking. It also times how long it takes to composite layers. Each layer is now given a name. The layer name is non optional when creating layers but can explicitly be set to "" to not have one .

I addded a config setting  <config/video/vsync> to control vsync. defaults to 1  but cab be set to anything SDL accepts. 0 disables vsync and is useful if testing performance and you want o know the maximum you can get. Not going to add this to the gump bcause the default 1 should be fine for everyone but people want to change it they can in exult.cfg

In the surface creation related changes I removed as much of the vestigial support for 8 bit display modes that still seemed to remain. 8 bit modes wouldn't work anyway with current exult so there was no reason to keep this code. I also remove the fallback unscaled surface creation in  Image_window::create_surface. Now if try_scaler fails when trying the point scaler fallback, it is a fatal error. I;m not sure of any scenario that would occur where try_scaler would fail and the fallback surface creation wouldn't also have problems, but I need to look at this some more in case there is a situation I've missed

I removed some Layer related code duplication in Text Gumps that codefactor complained about.

Still have a few things I want to do such as looking to speed up compositing of big layers like the Performance Metric Overlay. For me it is by far the slowest part of drawing a frame

When performance Metrics are being shown The Screen will be constantly redrawn to get an accurate idea of the framerate. For Slow systems this would be bad so Metrics overlay is off by default and must be turned on with Alt-2. Pressing Alt-2 once shows only fps and frame time. Pressing Alt-2  a second time shows the detailed metrics display. Pressing it a third time hides it.

This set of changes Requires SDL 3.2. I'm not sure if we check for a minimum version of SDL. I may need to update that 

@DominusExult can you test this  to see if scaling is working properly for you as this gets rid of your attempted fix of the SDLScaler problems that only seemed to affect you.

### DIFF
--- a/imagewin/iwin8.cc
+++ b/imagewin/iwin8.cc
@@ -326,8 +326,20 @@ bool Image_window8::refresh_layer_scaled(Layer& layer, int factor) {
 		const int    ossw  = ((logw + 3) & ~3) + 2 * gb;
 		const int    ossh  = logh + 2 * gb;
 		SDL_Surface* osrc8 = SDL_CreateSurface(ossw, ossh, SDL_PIXELFORMAT_INDEX8);
-		SDL_Surface* odst  = SDL_CreateSurface(ossw * factor, ossh * factor, SDL_PIXELFORMAT_ARGB8888);
-		bool         done  = false;
+		SDL_Surface* odst;
+		if (!SDL_LockTextureToSurface(layer.texture, nullptr, &odst)) {
+			const char* err = SDL_GetError();
+			std::cerr << "SDL_LockTextureToSurface failed trying to lock texture for layer " << layer.get_name() << ":"
+					  << (err ? err : "") << std::endl;
+
+			return false;
+		}
+		if (odst->w < ossw * factor || odst->h < ossh * factor) {
+			std::cerr << " odst surface size too small after locking texture for layer '" << layer.get_name() << "' expected "
+					  << (ossw * factor) << "x" << (ossh * factor) << " got " << odst->w << "x" << odst->h << std::endl;
+			return false;
+		}
+		bool done = false;
 		if (osrc8 && odst) {
 			SDL_Palette* sdl_pal = SDL_CreateSurfacePalette(osrc8);
 			if (sdl_pal) {
@@ -368,14 +380,12 @@ bool Image_window8::refresh_layer_scaled(Layer& layer, int factor) {
 							trow[x] = 0xff000000u | (row[x] & 0x00ffffffu);
 						}
 					}
-					SDL_UpdateTexture(layer.texture, nullptr, texpix.get(), tex_w * static_cast<int>(sizeof(uint32)));
+					SDL_UnlockTexture(layer.texture);
 					done = true;
 				}
 			}
 		}
-		if (odst) {
-			SDL_DestroySurface(odst);
-		}
+
 		if (osrc8) {
 			SDL_DestroySurface(osrc8);
 		}
--- a/imagewin/iwin8.cc
+++ b/imagewin/iwin8.cc
@@ -48,6 +48,7 @@ Boston, MA  02111-1307, USA.
 #include <algorithm>
 #include <climits>
 #include <cstring>
+#include <iostream>
 
 using std::make_unique;
 using std::rotate;
@@ -260,28 +261,37 @@ void Image_window8::refresh_layer(Layer& layer) {
 	if (layer.render_scale > 1 && refresh_layer_scaled(layer, layer.render_scale)) {
 		return;
 	}
-	const int            w      = layer.get_width();
-	const int            h      = layer.get_height();
-	Image_buffer*        b      = layer.get_ibuf();
-	const unsigned char* src    = b->get_bits();
-	const int            pitch  = static_cast<int>(b->get_line_width());
-	auto                 pixels = make_unique<uint32[]>(static_cast<size_t>(w) * h);
-	uint32               palette[256];
+	const int            w     = layer.get_width();
+	const int            h     = layer.get_height();
+	Image_buffer*        b     = layer.get_ibuf();
+	const unsigned char* src   = b->get_bits();
+	const int            pitch = static_cast<int>(b->get_line_width());
 
+	char* pixels       = nullptr;
+	int   pixels_pitch = 0;
+
+	if (!SDL_LockTexture(layer.texture, nullptr, reinterpret_cast<void**>(&pixels), &pixels_pitch)) {
+		const char* err = SDL_GetError();
+		std::cerr << "SDL_LockTexture failed trying to lock texture for layer " << layer.get_name() << ":" << (err ? err : "")
+				  << std::endl;
+		return;
+	}
+
+	uint32 palette[256];
 	for (int i = 0; i < 256; i++) {
 		palette[i] = layer_argb_pixel(layer, i);
 	}
 
 	for (int y = 0; y < h; y++) {
 		const unsigned char* srow = src + static_cast<size_t>(y) * pitch;
-		uint32*              drow = pixels.get() + static_cast<size_t>(y) * w;
+		uint32*              drow = reinterpret_cast<uint32*>(pixels + static_cast<size_t>(y) * pixels_pitch);
 		uint32*              end  = drow + w;
 
 		while (drow != end) {
 			*drow++ = palette[*srow++];
 		}
 	}
-	SDL_UpdateTexture(layer.texture, nullptr, pixels.get(), w * static_cast<int>(sizeof(uint32)));
+	SDL_UnlockTexture(layer.texture);
 }
 
 /*
--- a/imagewin/iwin8.cc
+++ b/imagewin/iwin8.cc
@@ -492,7 +492,16 @@ bool Image_window8::refresh_layer_scaled(Layer& layer, int factor) {
 		}
 	}
 
-	auto texpix = make_unique<uint32[]>(static_cast<size_t>(tex_w) * tex_h);
+	int     texpix_pitch;
+	char *texpix=nullptr;
+	if (!SDL_LockTexture(layer.texture, nullptr, reinterpret_cast<void**>(&texpix), &texpix_pitch)) {
+		const char* err = SDL_GetError();
+		std::cerr << "SDL_LockTexture failed trying to lock texture for layer " << layer.get_name() << ":"
+					<< (err ? err : "") << std::endl;
+
+		return false;
+	}
+
 	if (ok) {
 		const uint32* pix1             = static_cast<const uint32*>(dst32->pixels);
 		const uint32* pix2             = static_cast<const uint32*>(dst32b->pixels);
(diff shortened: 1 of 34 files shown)


DominusExult:
I'm running into several issues:
- Only scalers that seem to work correctly are point, interlaced, bilinear
- Other scalers at game area `Auto` a noticeable flicker again and again (see video link)
- Also there is a black guardband on the top and to the left 
- Other scalers at game area 320x200 -> black screen

This was with your unchanged fork. With my natural light fork I have the same issues but on top the light layer is not in sync with the game area (game area Auto, other scalers), probably due to the guardbands top/left.

Video that shows the flicker, blackness.
https://youtu.be/5xwYVUtMAko

wench:
can't replicate the black screen but others I seem to be able to replicate but the flickering seems to behave differently. On yours the flickering seems to be the screen shifting but for me the screen is resizing. Moving the mouse seems to affect it somehow. Feels like an uninitialized variable causing instability. Hopefully its all just the same problem. The flickering doesn't affect the visable guard bands so that seems to suggest they are not the same problem and gives me hints at where to look

DominusExult:
My SDL is not up to date, I just remembered. I'll try if the blackness is an SDL issue that has already been fixed

wench:
Going frame by frame in your video the flickering is also the screen resizing exactly as i see it, Good to know . I have some work to do. I thought this would be an easy fix

wench:
I seem to be wrong about the flickering not affecting the guardband, looking at more of your video frame by frame I can see it does. 

Could you try using SDL_RENDER_DRIVER=software

DominusExult:
1. SDL version is **not** the blackness culprit :(
2. using SDL_RENDER_DRIVER=software does fix the blackness with 320x200 and the guardband is also correct (not there in fill mode fill). It does **not** help with the flickering.
3. Yes, it is some resizing happening when moving the mouse. and it affects the guardband (at least the left one is a bit wider)
Before flicker:
image

Flicker frame:
image

wench:
I can see there are 2 problems in the code. Flickering occurs where there is a partial screen redraw. It stretches the redrawn dirty region over the entire screen. Turning on my perf metrics overlay causes all redraws to be of the entire screen so I never saw them as I was testing with the overlay enabled. The guardband problem is because the update region is not properly calculated when there is no inter_surface. The fix for both of these will be the same but requires some rethinking about what I'm doing. I've been making some assumptions about how SDL behaves but I'm pretty sure I'm wrong given the comment Lanica left in Image_window::UpdateRect() in 2013. Looking into it the SDL hardware accelerated 2d rendering support seems to be based around the idea that all updates to textures and the screen will be full updates. I've been assuming that a partial update to screen_texture would be ok, and with the way Direct3d works it probably works fine for me but would break for others.

wench:
I can see there are 2 problems in the code. Flickering occurs where there is a partial screen redraw. It stretches the redrawn dirty region over the entire screen. Turning on my perf metrics overlay causes all redraws to be of the entire screen so I never saw them as I was testing with the overlay enabled. The guardband problem is because the update region is not properly calculated when there is no inter_surface. The fix for both of these will be the same but requires some rethinking about what I'm doing. I've been making some assumptions about how SDL behaves but I'm pretty sure I'm wrong given the comment Lanica left in Image_window::UpdateRect() in 2013. Looking into it the SDL hardware accelerated 2d rendering support seems to be based around the idea that all updates to textures and the screen will be full updates. I've been assuming that a partial update to screen_texture would be ok, and with the way Direct3d works it probably works fine for me but would break for others.

The black screen problem Looks like an Alpha Channel Problem. I have managed to replicate black screen in 320x200 by making inter_surface ARGB. Inter surface is set to the same pixel format as the desktop display mode, For me it is XRGB. Not sure if this is the problem but looking at the SDL the Cocoa video code, The desktop display mode is always returned as ARGB where as the windows code always return XRGB. This is the only difference between MacOs and Windows that I can think is relevant and it replicates the behaviour. Now I need to decide what the correct fix is

Can I get you to try this 
Line 721 of Imagewin.cc add
`SDL_SetSurfaceBlendMode(inter_surface, SDL_BLENDMODE_NONE);`
after
```
		if (!(inter_surface = SDL_CreateSurface(i_width, i_height, desktop_displaymode.format))) {
			cerr << "Couldn't create inter surface: " << SDL_GetError() << endl;
			free_surface();
			return false;
		}
```

If I'm right that should fix the black screen. But I'm not sure if its the best solution

DominusExult:
Yes, this works.

Another thing I noticed is that the guardband is behaving strangely throughout the settings.
With window size 1440x810, scaler point x4 (or x3), game area auto, fill mode Fill and fill quality bilinear the guardband is there (not with point or SDLScaler) and with other scalers I had the guardband with point only (but I can't pin point with which it was).

wench:
I think all the problems should now be fixed

DominusExult:
Works great now, I think I tested all combination (though only with point and 4xbr main scaler).

One oddity, after extensively testing without restarting I enabled the FPS counter and had really low numbers (12-15 fps).
After testing some more I found that the fps dropped when using window size 1440x810, scaler point x1 and doing something with  **game area auto/320x200** and maybe fill quality and mode. I wasn't able to reproduce it reliably but something triggered the fps to drop low on game area auto and it didn't recover when setting back to 320x200. It happened once more but then I couldn't reproduce my steps. The location was Britains throne room/park at around 22:00.
Sorry, usually I can pin point the error sequence better :)

DominusExult:
One other bug I stumbled over:
The performance overlay acts up on macOS fullscreen. Very tiny compared to how it looks in windowed mode but only when I start windowed.
When I start in fullscreen it's even tinier and when I switch to windowed it stays tiny.
Same windowed and fullscreen settings (but it happens with all settings it seems):
1600x900
Point x8
Game Area 320x200
Fill qual Point
Fill mode fill

small part of the 320x200 windowed screen after starting in fullscreen, enabling the performance overlay, swithing back to windowed mode
image

Normal look:
image

wench:
I improved the behaviour of the Scaling of the performance text. the size of the text is Dependant on the display size.
At 1440 width the text is unscaled. At other sizes the text is scaled to maintain the same relative size with a minimum scale of .75 so the text doesn't become completely unreadable at small resolutions

I made a few optimizations to speed up the compositing perf text layer. I added a UILayerKind that always does no scaling except SDLScaler and I optimized Image_window8::refresh_layer

DominusExult:
I will not be able to test for the next ten days.

wench:
That's cool. No rush for this.
________________________________

wench:
Image_window8::refresh_layer_scaled() seems to create what I consider to be an excessive number of temporary SDL Surfaces. I plan on rewriting much of it to get rid of temporary surfaces and all the copying between surfaces.

DominusExult:
Tested and I have not noticed any issues now. I want to test it later with my natural light branch, though.

The natural light needed some fixes but other than that it works fine.
I left two comments where I'm not sure whether the guradband offset is not also needed in `void Image_window8::refresh_layer(Layer& layer)`

imagewin/iwin8.cc
```diff
@@ -260,20 +285,37 @@ void Image_window8::refresh_layer(Layer& layer) {
...
-	const int            w      = layer.get_width();
-	const int            h      = layer.get_height();
-	Image_buffer*        b      = layer.get_ibuf();
-	const unsigned char* src    = b->get_bits();
-	const int            pitch  = static_cast<int>(b->get_line_width());
-	auto                 pixels = make_unique<uint32[]>(static_cast<size_t>(w) * h);
+	const int            w            = layer.get_width();
+	const int            h            = layer.get_height();
+	SDL_Surface*         s            = layer.surface;
+	const unsigned char* src          = reinterpret_cast<unsigned char*>(s->pixels);
```
DominusExult: Doesn't this also need the offset for guardband and pitch?
`const unsigned char* src          = reinterpret_cast<unsigned char*>(s->pixels) + guard_band + static_cast<size_t>(guard_band) * pitch;`

imagewin/iwin8.cc
```diff
@@ -260,20 +285,37 @@ void Image_window8::refresh_layer(Layer& layer) {
...
+	for (int i = 0; i < 256; i++) {
+		palette[i] = layer_argb_pixel(layer, i);
+	}
+
 	for (int y = 0; y < h; y++) {
 		const unsigned char* srow = src + static_cast<size_t>(y) * pitch;
-		uint32*              drow = pixels.get() + static_cast<size_t>(y) * w;
-		for (int x = 0; x < w; x++) {
-			drow[x] = layer_argb_pixel(layer, srow[x]);
+		uint32*              drow = reinterpret_cast<uint32*>(pixels + static_cast<size_t>(y) * pixels_pitch);
```
DominusExult: And here as well?
`uint32* drow = reinterpret_cast<uint32*>(pixels + static_cast<size_t>(y + guard_band) * pixels_pitch) + guard_band;`

wench:
Those pointers are wrong, Thanks for the review. These would no doubt cause subtle difficult to find problems

DominusExult:
To make my natural light fork work, I needed to move the imagewin.cc `SDL_SetTextureBlendMode(layer.texture, layer.is_opaque() ? SDL_BLENDMODE_NONE : SDL_BLENDMODE_BLEND);` from line 2119 into the `if (layer.dirty) {` block at line 2124 (above `refresh_layer(layer);`) (with some additional bool for natural light).
Is that ok? And if so will you do it or should I do it when I merge my natural light fork (after you merge this)? 
Asking because I noticed your cloud fork also messes with blend mode.

wench:
Yeah I expect my cloud fork will have conflicts.  I'll deal. with those later I don't expect it will work well with the lighting changes Need to get the layer ordering right between the clouds and lights. Make whatever changes you need

DominusExult:
So ready to merge this?
Then I'll merge the natural lights next

wench:
Yeah I have nothing more to do here

